PR #114, opened 2021-06-10
remove dependency on feature(const_fn)

asclepiusaka:


asclepiusaka:
this might not be very back compatible, let's hold off on it.

PR #386, opened 2022-02-18
TCP over RDMA POC Wrap up

qingqu:


config.json
```diff
@@ -16,8 +16,8 @@
   "DedicateUring" : 0,
   "UringSize"     : 64,
   "UringEpollCtl" : true,
-  "EnableRDMA"    : false,
+  "EnableRDMA"    : true,
```
QuarkContainer, line 19: Please keep the default value

qvisor/src/vmspace/kernel_io_thread.rs
```diff
@@ -46,6 +47,7 @@ impl KIOThread {
     pub fn ProcessOnce(sharespace: &ShareSpace) -> usize {
         let mut count = 0;
 
+        count += RDMA.PollCompletionQueueAndProcess();
```
QuarkContainer, line 50: can you switch off ts when the RDMA is disabled?

QuarkContainer (changes requested):

PR #830, opened 2023-04-09
Log enhancement: enable qkernel capture log message from Libraries 

yaoxin1995:
The log crate provides a single logging API that abstracts over the actual logging implementation. Libraries use the logging API provided by this crate, and the consumer of those libraries (qkernel) uses the logging implementation `SimpleLogger` to capture and print the logs to `quark.log`

### DIFF
--- a/config.json
+++ b/config.json
@@ -18,7 +18,7 @@
   "RDMAPort"      : 1,
   "PerSandboxLog" : false,
   "ReserveCpuCount": 1,
-  "ShimMode"      : false,
+  "ShimMode"      : true,
   "EnableInotify" : true,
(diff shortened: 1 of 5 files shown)


config.json
```diff
@@ -18,7 +18,7 @@
   "RDMAPort"      : 1,
   "PerSandboxLog" : false,
   "ReserveCpuCount": 1,
-  "ShimMode"      : false,
```
QuarkContainer, line 21: Please keep the default ShimMode as false

QuarkContainer (changes requested):

config.json
yaoxin1995: done

PR #984, opened 2023-09-14
aarch64: HyperCall64 with MMIO write (ABI proposal, may need discussion, don't merge yet)

shrik3:
use `str` instruction to trigger MMIO_WRITE.  x0 and x1 are used as operands (addr and data). 
4 parameters are passed via register r2 ~ r5.

Perhaps need some discussions?  **Please don't merge yet**

### DIFF
--- a/qkernel/src/kernel_def.rs
+++ b/qkernel/src/kernel_def.rs
@@ -302,7 +302,25 @@ pub fn HyperCall64(type_: u16, para1: u64, para2: u64, para3: u64, para4: u64) {
 #[cfg(target_arch = "aarch64")]
 #[inline(always)]
 pub fn HyperCall64(type_: u16, para1: u64, para2: u64, para3: u64, para4: u64) {
-    // TODO add HyperCall64
+    // use MMIO to cause VM exit hence "hypercall"
+    // the data does not matter, 
+    // addr of the MMIO identifies the Hypercall.
+    // x0 and x1 (w1) are used by the str instruction
+    // the 64-bit parameters 1,2,3,4 are passed via
+    // x2,x3,x4,x5
+    unsafe {
+        let data: u8 = 0;
+        asm!("
+             str w1, [x0]
+             ",
+             in("x0") type_,
+             in("w1") data,
+             in("x2") para1,
+             in("x3") para2,
+             in("x4") para3,
+             in("x5") para4,
+             )
+    }
 }
 
 impl CPULocal {
--- a/qkernel/src/kernel_def.rs
+++ b/qkernel/src/kernel_def.rs
@@ -310,10 +310,12 @@ pub fn HyperCall64(type_: u16, para1: u64, para2: u64, para3: u64, para4: u64) {
     // x2,x3,x4,x5
     unsafe {
         let data: u8 = 0;
+        let addr: u64 = MemoryDef::HYPERCALL_MMIO_BASE + (type_ as u64);
+
         asm!("
              str w1, [x0]
              ",
-             in("x0") type_,
+             in("x0") addr,
              in("w1") data,
              in("x2") para1,
              in("x3") para2,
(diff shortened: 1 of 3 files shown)


abel-von:
image

This is what gvisor did, If this is better than what we do in this PR?

shrik3:
In [arm64-kvm-hello-world](https://github.com/Lenz-K/arm64-kvm-hello-world) there is an example of VMM allocating read-only memory to VM, so that writing to that region triggers MMIO_WRITE 

in vmm:
```cpp
/**
 * Copies the code into the memory of the specified memory mapping.
 *
 * @param code The code blok that will be copied into the VM memory.
 * @param memsz The size of the code block.
 * @param target_addr The VM memory address that the code will be copied to.
 * @param mmi The index of the memory mapping that will be used for copying.
 * @return Returns the index of the memory mapping or -1 if no mapping was found.
 */
int copy_section_into_memory(uint32_t *code, size_t memsz, uint64_t target_addr, int mmi) {
    // There can be an offset between memory mapping and the target address.
    uint64_t offset = target_addr - memory_mappings[mmi].guest_phys_addr;

    // If the offset plus the code size is bigger than the memory mapping size, do nothing.
    if (offset + memsz > memory_mappings[mmi].memory_size) {
        printf("Memory mapping too small. Mapping offset: 0x%08lX - Mapping size: 0x%08lX\n", offset, memory_mappings[mmi].memory_size);
        return -1;
    }

    // Copy the code into the VM memory
    memcpy(memory_mappings[mmi].userspace_addr + offset, code, memsz);
    printf("Section loaded. Host address: %p - Guest address: 0x%08lX\n", memory_mappings[mmi].userspace_addr + offset, target_addr);
    return 0;
}
```
and 

```
check_vm_extension(KVM_CAP_READONLY_MEM, "KVM_CAP_READONLY_MEM");
// This will cause a write to 0x10000000, to result in a KVM_EXIT_MMIO.
mem = allocate_memory_to_vm(MEMORY_BLOCK_SIZE, 0x10000000, KVM_MEM_READONLY);
```



I would say this is more flexible than gvisor's approach.

qkernel/src/kernel_def.rs
```diff
@@ -302,7 +302,27 @@ pub fn HyperCall64(type_: u16, para1: u64, para2: u64, para3: u64, para4: u64) {
...
+    // addr of the MMIO identifies the Hypercall.
+    // x0 and x1 (w1) are used by the str instruction
+    // the 64-bit parameters 1,2,3,4 are passed via
+    // x2,x3,x4,x5
+    unsafe {
+        let data: u8 = 0;
+        let addr: u64 = MemoryDef::HYPERCALL_MMIO_BASE + (type_ as u64);
+
+        asm!("
+             str w1, [x0]
```
CharlyYu, line 316: should using immediate number be better than loading 0 into this register w1？

CharlyYu:
But we didn't mmap and set memory region of the address HYPERCALL_MMIO_BASE,  and is it a waste of memory if we mmap and set this memory to vm?

qkernel/src/kernel_def.rs
shrik3: I thought about that, but I don't think you can use immv as the first operand of `str`

https://developer.arm.com/documentation/dui0489/h/arm-and-thumb-instructions/ldr-and-str--immediate-offset-

in all cases the first operand is a register.

shrik3:
> But we didn't mmap and set memory region of the address HYPERCALL_MMIO_BASE

Yes, with this approach we need to also modify the qvisor code accordingly.

> and is it a waste of memory if we mmap and set this memory to vm?

a 16 bit (64K) address space would be more than enough for hypercalls,  also since it's virtual memory,  I don't think it's a waste.

And the number can be lower: with only a dozen of hypercalls, a 8 bit address space would be enough.

shrik3:
This is merged with #1000 . Close.

PR #1005, opened 2023-10-15
aarch64: add vdso of arm64

CharlyYu:
1. add vdso compile script to get a vdso of arm64
2. load vdso and init timekeeper with the new vdso
3. annotate vsyscall related codes with x86_64 as it is not neccesary for aarch64


qlib/mod.rs
```diff
@@ -126,26 +126,47 @@ pub fn InitSingleton() {
     }
 }
 
-pub const HYPERCALL_PANIC: u16 = 2;
```
shrik3, line 129: it was tested that 8 byte alignment for MMIO addr is not necessary when MMU is enabled.

qkernel/src/lib.rs
```diff
@@ -442,43 +442,28 @@ pub extern "C" fn rust_main(
...
-        Kernel::HostSpace::SyncPrint(self::qlib::config::DebugLevel::Debug, msg2);
-        raw!(0x240, id, vcpuCnt, 2);
         GLOBAL_ALLOCATOR.Init(heapStart);
-        //let msg = format!("[INFO] cpu{} {}", id, "init sharespace");
-        Kernel::HostSpace::SyncPrint(self::qlib::config::DebugLevel::Trace, "init sharespace");
-        raw!(0x240, id, vcpuCnt, 3);
         SHARESPACE.SetValue(shareSpaceAddr);
         SingletonInit();
-
+        error!("init singleton finished");
```
chl337, line 450: For clarity (and consistency), I suggest further usage of `debug!(...)` or `info!(...)`.

qlib/mod.rs
```diff
@@ -126,26 +126,47 @@ pub fn InitSingleton() {
...
+//pub const HYPERCALL_HCALL: u16 = 15;
+//pub const HYPERCALL_IOWAIT: u16 = 16;
+//pub const HYPERCALL_WAKEUP_VCPU: u16 = 17;
+//pub const HYPERCALL_EXIT_VM: u16 = 18;
+//pub const HYPERCALL_VCPU_FREQ: u16 = 19;
+//pub const HYPERCALL_VCPU_YIELD: u16 = 20;
+//pub const HYPERCALL_VCPU_DEBUG: u16 = 21;
+//pub const HYPERCALL_VCPU_PRINT: u16 = 22;
+//pub const HYPERCALL_VCPU_WAIT: u16 = 23;
+//pub const HYPERCALL_RELEASE_VCPU: u16 = 24;
```
chl337, line 148: If there is no further need for this lines, it will be better to remove them at all.

qkernel/src/lib.rs
abel-von: I thought this is just a temporary change, and we have to remove it, and yes I will change it to debug

qlib/kernel/asm/aarch64.rs
```diff
@@ -99,7 +99,11 @@ pub fn IRet(kernelRsp: u64) -> ! {
 
 #[inline]
 pub fn GetRsp() -> u64 {
-    return 0;
+    let rsp: u64;
+    unsafe {
+        asm!("mov {0}, sp", out(reg) rsp);
```
SchwarzStein, line 104: could you please clarify the usage of GetRsp.
In aarch64, sp can be both userspace and kernel space stack pointer.
If for all instances here you need to have the kernel one, please use sp_el1 otherwise, 
it should be okay.

qvisor/src/runc/runtime/vm.rs
```diff
@@ -370,8 +370,8 @@ impl VirtualMachine {
         info!("before loadKernel");
 
         let entry = elf.LoadKernel(Self::KERNEL_IMAGE)?;
-        //elf.LoadVDSO(&"/usr/local/bin/vdso.so".to_string())?;
-        //VMS.lock().vdsoAddr = elf.vdsoStart;
+        elf.LoadVDSO(&"/usr/local/bin/vdso.so".to_string())?;
```
SchwarzStein, line 373: this seems to be an arbitrary choice for the path to vdso.so.
Could we have it as a global constant or configurable option.
I guess, if we have inherited this from x86, it would be sufficient to have
it defined as a global constant. This way, we can refactor the code as we
introduce new features.

vdso/makefile
```diff
@@ -1,21 +1,25 @@
+arch = $(shell uname -m)
```
SchwarzStein, line 1: I think this is incorrect. It would not work with cross-compilation as the target would be native.
i.e. if you are on x86 and are cross-compiling to arm, this would cause it to compile to arm??
Please, prefer to use environment exported by the build environment.
e.g. CARGO_CFG_TARGET_ARCH

vdso/makefile
```diff
@@ -1,21 +1,25 @@
...
+LINKER_SCRIPT = vdso_$(arch).lds
+
 .PHONY: all
 all: $(OUTPUTFILE)
 
 $(OUTPUTFILE): $(subst .cc,.o,$(SOURCES))
 	$(CC) $(CCFLAGS) -I. \
 	-O2 -std=c++11 \
-	-fPIC -fno-stack-protector -fuse-ld=gold -m64 -shared -nostdlib \
+	-fPIC -fno-stack-protector -fuse-ld=gold -shared -nostdlib \
```
SchwarzStein, line 18: is there a reason to drop the -m64 flag?

qlib/kernel/asm/aarch64.rs
shrik3: the same goes for functions like `pub fn Rdtsc()` -- arm doesn't have `rdtsc` in the first place! but we are keeping the naming to be consistent with existing code.  To resolve this we need refactoring with another indirection.

vdso/makefile
chl337: "-m64" seems to not be a valid option for Aarch64 according to the [GCC - index](https://gcc.gnu.org/onlinedocs/gcc/Option-Index.html).

qlib/kernel/asm/aarch64.rs
SchwarzStein: My comment regards functional correctness, not naming consistency.
If, for instance, in an exception situation you are reading the sp, what would the code be expecting, the kernel stack or the userspace stack. With this consideration, you may need to configure for aarch64 to ensure that, while chasing consistent abstraction, functional correctness is maintained.

vdso/makefile
CharlyYu: There is no such env here because vdso is not build by cargo, and I think the makefile in Quark also detect arch by this, and we can consider support cross-compilation later in a unified way.

CharlyYu:
> Otherwise we need to remove *.o ,vdso.so and vdso.d so that vdso can be built for aarch64 out of the box.
yes, it it removed in this PR now.

qvisor/src/runc/runtime/vm.rs
CharlyYu: done

qlib/kernel/asm/aarch64.rs
CharlyYu: Yes actually there are many codes that need to be reconstructed, but we are trying to build a runnable Quark on aarh64 so the clean codes is not yet in our consideration and I think we can make some big reconstruction of codes after a runnable Quark is ready.

CharlyYu: @SchwarzStein @chl337  I have submit another PR to rename the GetRsp to GetCurrentKernelSp and implement it in aarch64, although we still have to get sp_el1 by `mov {0} sp`, we set the SPSel to 1 before we do it. #1008 

And the reason we don't use `mrs` is that we can only mrs sp_el1 in EL2 and EL3 according to this webpage https://forums.raspberrypi.com/viewtopic.php?t=226080

vdso/makefile
SchwarzStein: we will add cross compile features after, as you suggested.

PR #1007, opened 2023-10-21
Support vcpu local variable 

CharlyYu:
store vcpu local variable into tpidr_el1

PR #1008, opened 2023-10-23
aarch64: rename GetRsp to GetCurrentKernelSp and implement it in aarch64

CharlyYu:
NOTE: we can only get sp_el1 in EL2 and EL3, so we have to get sp_el1 by mov and set the SPSel

### DIFF
--- a/qlib/kernel/asm/aarch64.rs
+++ b/qlib/kernel/asm/aarch64.rs
@@ -38,10 +38,11 @@ pub fn SwapGs() {
 
 pub fn GetVcpuId() -> usize {
     let ret: u64;
-    let tls = unsafe { tpidr_el1() };
     unsafe {
-        asm!("ldr {1}, [{0}, #16]",
-        in(reg) tls,
+        asm!("\
+        mrs {0}, tpidr_el1
+        ldr {1}, [{0}, #16]",
+        out(reg) _,
         out(reg) ret);
     };
     return ret as usize;
@@ -107,19 +108,31 @@ pub fn IRet(kernelRsp: u64) -> ! {
 }
 
 #[inline]
-pub fn GetRsp() -> u64 {
-    let rsp: u64;
+pub fn GetCurrentKernelSp() -> u64 {
+    // we can only mrs sp_el1 in EL2 and EL3
+    // so we can only get sp_el1 by move
+    let ret: u64;
     unsafe {
-        asm!("mov {0}, sp", out(reg) rsp);
-    };
-    return rsp;
+        asm!("\
+        mrs {1}, spsel
+        msr spsel, #1
+        mov {0}, sp
+        msr spsel, {1}
+        ",
+        out(reg) ret,
+        out(reg) _);
+    }
+    ret
+}
+
+#[inline]
+pub fn GetCurrentUserSp() -> u64 {
+    unsafe { return sp_el0(); }
 }
 
 #[inline]
 pub fn Clflush(addr: u64) {
 }
-
-
 
 // HostID executes a native CPUID instruction.
 // return (ax, bx, cx, dx)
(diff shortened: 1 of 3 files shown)


shrik3:
Is #1007   ready for a review?  I see the commits from the other two PRs are also included here

qlib/kernel/asm/aarch64.rs
```diff
@@ -107,19 +107,30 @@ pub fn IRet(kernelRsp: u64) -> ! {
...
+pub fn GetCurrentKernelSp() -> u64 {
+    // we can only mrs sp_el1 in EL2 and EL3
+    // so we can only get sp_el1 by move
+    let ret: u64;
     unsafe {
-        asm!("mov {0}, sp", out(reg) rsp);
-    };
-    return rsp;
+        asm!("\
+        mrs x20, spsel
```
shrik3, line 116: If I understand correctly x20 here is used to save the original value of spsel.
However I think this may breaks the calling convention because x20 is callee saved.

To explicitly tell the compiler to "clobber" the register, maybe add some thing like:
```rust
out("x20") _,
```

qlib/kernel/asm/aarch64.rs
```diff
@@ -524,3 +535,30 @@ pub unsafe fn midr() -> u64 {
...
+    let ret: u64;
+    asm!("mrs {0}, sp_el0", out(reg) ret);
+    ret
+}
+
+pub unsafe fn sp_el0_write(val: u64) {
+    asm!("msr sp_el0, {0}", in(reg) val);
+}
+
+pub unsafe fn sp_el1() -> u64 {
```
shrik3, line 550: since reading/writing SP_EL1 via MRS/MSR is illegal in EL1, and qkernel does not go beyond EL1, perhaps we can remove the function `sp_el1` and `ap_el1_write` since they won't work?

qlib/kernel/asm/aarch64.rs
```diff
@@ -524,3 +535,30 @@ pub unsafe fn midr() -> u64 {
     asm!("mrs {0}, midr_el1", out(reg) ret);
     ret
 }
+
+
+pub unsafe fn sp_el0() -> u64 {
```
chl337, line 540: 1. This function should be annotated with `[inline]`, to be on the safe side that the compiler will not create a new frame.
2. Is this nested call actually needed? `asm!("...", ...);` could be placed directly inside `pub fn GetCurrentUserSp() -> u64`

qlib/kernel/asm/aarch64.rs
CharlyYu: yes, you are right, and modified to use a scratch register

CharlyYu:
>Is https://github.com/QuarkContainer/Quark/pull/1007 ready for a review? I see the commits from the other two PRs are also included here

I just want to submit a pr to explain the modification about the comment in #1005, and after #1005 and #1007 is merged, I will rebase this commit.

qlib/kernel/asm/aarch64.rs
CharlyYu: Actually there is no place to call the GetCurrentUserSp until now, And I think we can make all functions of the read of a register with it's name and write of the register with the form of `<register_name>_write`, so that it is in such a unified way. The GetCurrentUserSp is a function name to let the callers knows what they are doing. 

And I think as the function is annotated inline, compiler will do the optimization so we don't have to care about the overhead of extra function frame.

CharlyYu: > This function should be annotated with [inline], to be on the safe side that the compiler will not create a new frame.

added

qlib/kernel/asm/aarch64.rs
CharlyYu: removed

SchwarzStein:
please rebase on the current armdev for this request to continue ...

SchwarzStein:
could you please make the functions you have exported as pub unsafe in aarch64.rs to pub <fn> and move the unsafe
keyword to wrap the register read/write. Since we are not dereferencing memory here, and are just working with primitive types.

CharlyYu:
> please rebase on the current armdev for this request to continue ...

done

CharlyYu:
> could you please make the functions you have exported as pub unsafe in aarch64.rs to pub and move the unsafe keyword to wrap the register read/write. Since we are not dereferencing memory here, and are just working with primitive types.

done

CharlyYu:
Is it ok to merge this PR? @SchwarzStein @shrik3 @chl337

PR #1009, opened 2023-10-27
[RFC] Add exception table and syscall handler

shrik3:
This PR adds the exception vector table and syscall handler.

I haven't tested this implementation (at least code compiles and runs) because we can't go to userspace yet, and only the el0->el1 exceptions are handled.

**Revision(s)**
- v1 > v2 : fixed typo, remove unnecessary `mut`
- v2 > v3: rebase against armdev with #1011  ; pass NULL as call frame pointer upon empty entry.
- v3 > v4: refined register save/restore for el0 and el1 exceptions; include sp_{el0,el1} into the frame. Renamed ExceptionStateEl1 to ExceptionFrame.

Major update (v4 > v5) :

- unified exception frame and PtRegs definitions.
- added diagnostics for data/instr aborts.
- added exception handlers for el1h_sync

small fix: fixed DFSC_ names (should be _Ln instead of _ELn)

Also please note that since we use PtRegs now (which is added in a newer PR #1027), this PR only work in combination of that one. (PtRegs for aarch64 undefined, otherwise). 

The pagefault handling is out of the scope of this PR. I suggest we merge this one first if there are no other issues.

### DIFF
--- a/qkernel/src/interrupt/aarch64/mod.rs
+++ b/qkernel/src/interrupt/aarch64/mod.rs
@@ -1,6 +1,309 @@
+use super::super::syscall_dispatch_aarch64;
+use crate::qlib::linux_def::MmapProt;
+use crate::qlib::addr::AccessType;
+use crate::qlib::kernel::SignalDef::PtRegs;
 pub unsafe fn InitSingleton() {
 }
 
 pub fn init() {
-    // TODO load IDT
+    // TODO set up GIC
+}
+
+use core::arch::asm;
+
+
+pub struct EsrDefs{}
+pub struct ISSDataAbort {}
+pub struct ISSInstrAbort {}
+
+// TODO move the defs to another module if it's too big.
+impl EsrDefs{
+    // ESR defs for Exception Class
+    pub const ESR_IL:u64            = 0x01 << 25;
+    pub const ESR_EC_SHIFT:u64      = 26;
+    pub const ESR_EC_MASK:u64       = 0x3f << 26;
+    pub const EC_UNKNOWN:u64        = 0x00;    /* Unkwn exception */
+    pub const EC_FP_SIMD:u64        = 0x07;    /* FP/SIMD trap */
+    pub const EC_BRANCH_TGT:u64     = 0x0d;    /* Branch target exception */
+    pub const EC_ILL_STATE:u64      = 0x0e;    /* Illegal execution state */
+    pub const EC_SVC:u64            = 0x15;    /* SVC trap */
+    pub const EC_MSR:u64            = 0x18;    /* MSR/MRS trap */
+    pub const EC_FPAC:u64           = 0x1c;    /* Faulting PAC trap */
+    pub const EC_INSN_ABORT_L:u64   = 0x20;    /* Instruction abort, from lower EL */
+    pub const EC_INSN_ABORT:u64     = 0x21;    /* Instruction abort, from same EL */ 
+    pub const EC_PC_ALIGN:u64       = 0x22;    /* PC alignment fault */
+    pub const EC_DATA_ABORT_L:u64   = 0x24;    /* Data abort, from lower EL */
+    pub const EC_DATA_ABORT:u64     = 0x25;    /* Data abort, from same EL */ 
+    pub const EC_SP_ALIGN:u64       = 0x26;    /* SP alignment fault */
+    pub const EC_TRAP_FP:u64        = 0x2c;    /* Trapped FP exception */
+    pub const EC_SERROR:u64         = 0x2f;    /* SError interrupt */
+    pub const EC_SOFTSTP_EL0:u64    = 0x32;    /* Software Step, from lower EL */
+    pub const EC_SOFTSTP_EL1:u64    = 0x33;    /* Software Step, from same EL */
+    pub const EC_WATCHPT_EL1:u64    = 0x35;    /* Watchpoint, from same EL */
+    pub const EC_BRK:u64            = 0x3c;    /* Breakpoint */
+
+    pub const ESR_CM:u64            = 0x01 << 8;     /* Cache Maintenance */
+    pub const ESR_WNR:u64           = 0x01 << 6;     /* Write not read */
+
+    // ESR defs for ISS
+    pub const ISS_FNV:u64       = 0x01 << 10;      /* FAR not valid */
+    pub const ISS_DFSC_MASK:u64 = 0x3f << 0;    /* DFSC and IFSC masks are the same */
+    // Access Size Fault L0~3
+    pub const DFSC_ASF_L0:u64  = 0x0;
+    pub const DFSC_ASF_L1:u64  = 0x1;
+    pub const DFSC_ASF_L2:u64  = 0x2;
+    pub const DFSC_ASF_L3:u64  = 0x3;
+
+    // Translation Fault L0~3
+    pub const DFSC_TF_L0:u64   = 0x4;
+    pub const DFSC_TF_L1:u64   = 0x5;
+    pub const DFSC_TF_L2:u64   = 0x6;
+    pub const DFSC_TF_L3:u64   = 0x7;
+
+    // Access Flag Fault L0~3
+    pub const DFSC_AF_L0:u64   = 0x8;      /* if FEAT_LPA2 is implmented */
+    pub const DFSC_AF_L1:u64   = 0x9;
+    pub const DFSC_AF_L2:u64   = 0xa;
+    pub const DFSC_AF_L3:u64   = 0xb;
+
+    // Permission Fault L0 ~ 3
+    pub const DFSC_PF_L0:u64   = 0xc;      /* if FEAT_LPA2 is implemented */
+    pub const DFSC_PF_L1:u64   = 0xd;
+    pub const DFSC_PF_L2:u64   = 0xe;
+    pub const DFSC_PF_L3:u64   = 0xf;
+
+    // Synchronous External Abort (SEA)
+    // not on translation table walk
+    pub const DFSC_SEA:u64      = 0x10;
+    // on translation table walk, L -1~3
+    pub const DFSC_SEA_M1:u64   = 0x13;     /* if FEAT_LPA2 is implemented*/
+    pub const DFSC_SEA_L0:u64  = 0x14;
+    pub const DFSC_SEA_L1:u64  = 0x15;
+    pub const DFSC_SEA_L2:u64  = 0x16;
+    pub const DFSC_SEA_L3:u64  = 0x17;
+
+    // Synchronous parity or ECC error when FEAT_RAS NOT implemented.
+    // not on table walk
+    pub const DFSC_ECC:u64      = 0x18;
+    // on table walk: L-1~3
+    pub const DFSC_ECC_M1:u64   = 0x1b;
+    pub const DFSC_ECC_L0:u64  = 0x1c;
+    pub const DFSC_ECC_L1:u64  = 0x1d;
+    pub const DFSC_ECC_L2:u64  = 0x1e;
+    pub const DFSC_ECC_L3:u64  = 0x1f;
+
+    pub const DFSC_ALIGN:u64    = 0x21;
+
+    // the Granule Protection Faults
+    // GPF on table walk, L-1~3
+    pub const DFSC_GPF_M1:u64   = 0x23;
+    pub const DFSC_GPF_L0:u64  = 0x23;
+    pub const DFSC_GPF_L1:u64  = 0x25;
+    pub const DFSC_GPF_L2:u64  = 0x26;
+    pub const DFSC_GPF_L3:u64  = 0x27;
+    // GPF not on table walk
+    pub const DFSC_GPF:u64      = 0x28;
+
+    pub const DFSC_ASF_M1:u64   = 0x29;
+    pub const DFSC_TF_M1:u64    = 0x2b;
+
+    pub const DFSC_TLB_CONFLICT:u64 = 0x30;
+    // unsupported atomic hardware update
+    pub const DFSC_UAHU:u64     = 0x31;
+
+    #[inline]
+    pub fn GetExceptionFromESR(esr:u64) -> u64{
+        return (esr & EsrDefs::ESR_EC_MASK) >> EsrDefs::ESR_EC_SHIFT;
+    }
+
+    #[inline]
+    pub fn IsCM(esr:u64) -> bool {
+        return (esr & EsrDefs::ESR_CM) != 0;
+    }
+
+    #[inline]
+    pub fn IsWnR(esr:u64) -> bool {
+        return (esr & EsrDefs::ESR_WNR) != 0;
+    }
+}
+
+pub fn GetEsrEL1() -> u64 {
+    unsafe {
+        let value:u64;
+        asm!(
+            "mrs  {}, esr_el1",
+            out(reg) value,
+           );
+        return value;
+    }
+}
+
+// TODO, the FAR is only valid when ESR.ISS.FnV==0
+// For faults other than data/instr aborts, this flag
+// should be checked before using FAR.
+pub fn GetFarEL1() -> u64 {
+    unsafe {
+        let value:u64;
+        asm!(
+            "mrs  {}, far_el1",
+            out(reg) value,
+           );
+        return value;
+    }
+}
+
+pub fn GetException() -> u64{
+    let esr:u64 = GetEsrEL1();
+    return EsrDefs::GetExceptionFromESR(esr);
+}
+
+
+
+#[no_mangle]
+pub extern "C" fn exception_handler_unhandled(_ptregs_addr:usize, exception_type:usize){
+    // MUST CHECK ptr!=0 before dereferencing
+    // ptr == 0 indicates an empty entry in the exception table,
+    // in which case the context won't be saved/restored by the wrapper
+    // and this function MUST NOT return
+    panic!("unhandled exception - {}",
+           match exception_type {
+               0 => "EL1T_SYN",
+               1 => "EL1T_IRQ",
+               2 => "EL1T_FIQ",
+               3 => "EL1T_SERROE",
+               4 => "EL1H_SYN",
+               5 => "EL1H_IRQ",
+               6 => "EL1H_FIQ",
+               7 => "EL1H_SERROE",
+               _ => "NON-DEFINED",
+          }
+    );
+}
+
+#[no_mangle]
+pub extern "C" fn exception_handler_el1h_sync(ptregs_addr:usize){
+    let esr = GetEsrEL1();
+    let ec = EsrDefs::GetExceptionFromESR(esr);
+
+    match ec {
+        EsrDefs::EC_DATA_ABORT => {
+            let far = GetFarEL1();
+            MemAbortKernel(ptregs_addr, esr, far, false);
+        },
+        EsrDefs::EC_INSN_ABORT => {
+            let far = GetFarEL1();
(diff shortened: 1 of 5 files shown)


qkernel/src/lib.rs
```diff
@@ -332,6 +330,113 @@ pub extern "C" fn syscall_handler(
...
 }
 
+// syscall_handler implementation for aarch64:
+// Unlike x86, this function is NOT directly called 
+// from the asm code (vector). The C calling convention
+// is not necessary here.
+// TODO move this function to a proper place.
+#[no_mangle]
+#[cfg(target_arch = "aarch64")]
+pub fn syscall_dispatch_aarch64(
```
shrik3, line 340: Explanation:

I think this "dispatch" function is perhaps defined in a wrong place, but I'm trying to keep the flow similar to the x86 version (the `syscall_handler` function defined in lib.rs).

qkernel/src/interrupt/aarch64/mod.rs
```diff
@@ -1,6 +1,137 @@
+use super::super::syscall_dispatch_aarch64;
```
shrik3, line 1: It's maybe better to move this to somewhere like /qlib/kernel/arch/aarch64/exception.rs

However I'm having some issues with module resolution. I have no idea why that happens:  when I put this file under qlib,  importing `syscall_dispatch_aarch64` from lib.rs causes compilation issue. Weirdly, `cargo check` has no error but `cargo build` does.

The only way I can make it work is to define this module directly in qkernel's mod tree, instead of under qlib. Any suggestion?

shrik3 (commented):
-

qkernel/src/interrupt/aarch64/mod.rs
chl337: What was path-chain in "lib.rs"?

To go a bit further, in my opinion naming/location is in general a deeper structural challenge/issue, e.g. in qlib/:

1. For easing porting the asm/aarch64.rs mirrors function(instriction)-names from asm/x86_64.rs.
2. qlib seems "overloaded", if an element is only used by either QKernel or QVisor then should be placed in either in qkernel/ or qvisor/.

shrik3: > What was path-chain in "lib.rs"?

I'm not sure what you mean by "path-chain" but here is a failed attempt to import from lib.rs

I tried to put the exception handler  in `/qkernel/src/qlib/kernel/arch/aarch64/exception.rs`

and do
```rust
use super::super::super::super::super::syscall_dispatch_aarch64;
// I can't recall if I used exactly 5x supers but I was certainly referring to the crate root.
```

I'll get error like "can't find syscall_..... in crate root".  But when I put the exception.rs in qkernel/interrupts/aarch64/mod.rs like I'm doing in this PR, everything works fine.

> qlib seems "overloaded" ......

I totally agree! I'm tempted to put the architectural bits under /qkernel . IMHO the qlib should only serve as an interface between the two parts.

qkernel/src/interrupt/aarch64/mod.rs
```diff
@@ -1,6 +1,137 @@
...
+}
+
+use core::arch::asm;
+
+#[repr(C)]
+pub struct ExceptionStateEL1{
+    pub spsr_el1: u64,
+    pub elr_el1: u64,
+    pub esr_el1: u64,
+    pub tpidr_1l0: u64,
```
chl337, line 16: "_1l0" -> maybe typo here?

shrik3: yes it is, I'll fix it in the next revision.

qlib/kernel/arch/aarch64/exception.s
```diff
@@ -0,0 +1,178 @@
...
+    ldp x26, x1, [sp], #16
+    ldp x28, x1, [sp], #16
+    ldp x30, xzr, [sp], #16
+.endm
+
+// mitigaton of specter bhi see
+// TODO insert mitigation to the handler flow if the exception is taken
+// from EL0
+// https://documentation-service.arm.com/static/623c60d13b9f553dde8fd8e6?token=
+.macro spectre_bhb_loop cnt
```
chl337, line 72: Seems to not be used. (In the cases exception is _taken from_ EL0)

shrik3: yes I intentionally leave out the mitigation for the sake of simplicity (since the basic routine is not yet tested)

I mean the mitigation itself is trivial: adding a simple loop before EL0->1 transition does the job, however the cnt parameter in the mitigation should be configurable i.e. we need to define multiple entry points (and even multiple exception tables)[^multiple] with different loop counts, which will largely increase the code complexity and could make it hard to debug.

We can add BHI mitigation when the basic syscall routine is tested. 

[^multiple]: both linux an the BSD kernel do this. BSD kernel also switches back and forth between two disjoint vector tables upon enter/return. I suggest we simply ignore this complexity and first make things work.

qkernel/src/interrupt/aarch64/mod.rs
chl337:
```
use super::super::super::super::super::syscall_dispatch_aarch64;
// I can't recall if I used exactly 5x supers but I was certainly referring to the crate root.```suggestion
```
As we already discussed this in the "real world", we can mark it as _resolved_.

qkernel/src/interrupt/aarch64/mod.rs
```diff
@@ -1,6 +1,137 @@
...
+    pub const EC_WATCHPT_EL1:u64    = 0x35;    /* Watchpoint, from same EL */
+    pub const EC_BRK:u64            = 0x3c;    /* Breakpoint */
+    pub fn GetExceptionFromESR(esr:u64) -> u64{
+        return (esr & EsrDefs::ESR_EC_MASK) >> EsrDefs::ESR_EC_SHIFT;
+    }
+}
+
+pub fn GetEsrEL1() -> u64 {
+    unsafe {
+        let mut value:u64;
```
chl337, line 53: The _mut_ is redundant here.

qlib/kernel/arch/aarch64/exception.s
```diff
@@ -0,0 +1,178 @@
...
+
+.macro v_entry elx handler
+.align 7
+    b \handler
+.endm
+
+
+
+.align 11
+.globl vector_table
```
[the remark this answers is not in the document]
chl337: Yes, @shrik3 will you take care of it?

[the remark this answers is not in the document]
shrik3: No, it's syntactically correct. 

But for consistency I can switch to `global`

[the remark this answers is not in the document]
shrik3: I was about to change it into global but then I realize the existing x86 bits are using `globl` (see syscall_entry.s).

Perhaps I should keep it this way?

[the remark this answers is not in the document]
chl337: Yeap, my bad! I had the impression that it was ```global```.

qlib/kernel/arch/aarch64/exception.s
```diff
@@ -0,0 +1,178 @@
...
+vector_table:
+//          ELx    Handler
+v_empty     1      // we don't use SP_EL0 in EL1
+v_empty     1
+v_empty     1
+v_empty     1
+
+// interrupts are currently masked kernel
+v_entry     1      enter_el1h_sync
+v_empty     1      // TODO add entries
```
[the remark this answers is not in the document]
shrik3: I already did:

macro `v_empty`  wraps around a handler that cause panic:

```
.macro v_empty elx
.align 7
    bl exception_handler_unhandled
.endm
```
and 

```
pub extern "C" fn exception_handler_unhandled(_exception_frame_ptr:usize){
    panic!("unhandled exception");
}
```

But I forgot to pass the call frame ptr as parameter (perhaps we don't need it at all but I'll add)

qlib/kernel/arch/aarch64/syscall_entry.s
```diff
@@ -1,8 +1,10 @@
-.globl syscall_entry, CopyPageUnsafe
+.globl CopyPageUnsafe
 .globl context_swap, __vsyscall_page
 
-syscall_entry:
-    ret
+// For aarch64 this file does not serve as the actuall syscall entry
+// The actuall syscall is caused by SVC exception and the entry is vector_table
+// with offset 0x200, see exception.s and exception.rs
+// TODO rename this file
```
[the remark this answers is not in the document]
shrik3: I'd like to leave this issue open for now. Because refactoring `syscall_entry` would conflict with other ongoing works.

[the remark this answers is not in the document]
SchwarzStein: which work are you referring to ?

qkernel/src/interrupt/aarch64/mod.rs
```diff
@@ -1,6 +1,141 @@
...
-    // TODO load IDT
-}
\ No newline at end of file
+    // TODO set up GIC
+}
+
+use core::arch::asm;
+
+#[repr(C)]
+pub struct ExceptionStateEL1{
```
CharlyYu, line 12: Could you please give some explainations of why these registers  are saved?

qkernel/src/interrupt/aarch64/mod.rs
```diff
@@ -1,6 +1,141 @@
...
+    pub spsr_el1: u64,
+    pub elr_el1: u64,
+    pub esr_el1: u64,
+    pub tpidr_el0: u64,
+    pub x: [u64;32]
```
CharlyYu, line 17: Why are there 32 general registers? isn't it 31? I don't see where to save x31 in `el0_sync_save_regs`, and we don't save sp because we handle exception on user stack?

shrik3:
@CharlyYu 

> Why are there 32 general registers? isn't it 31?

There are 31, the extra is for padding, because the "push register" instruction stp stores in pair.
I can make the padding explicit like below, if this is preferred.
```
pub x: [u64;31];
 _res: u64;
```
In any case, pushing an odd number of registers without padding is not applicable.

> I don't see where to save x31

In the same sense, there is no x31 being saved, It is [x30 being saved twice](https://github.com/QuarkContainer/Quark/blob/e90b4a56aa9b2c17ac5fe1515c92b6c4e71a5b06/qlib/kernel/arch/aarch64/exception.s#L17) for the sole purpose of padding.

> and we don't save sp because we handle exception on user stack?

No, the kernel and user use different stack i.e. sp_el1 and sp_el0. The CPU automatically changes the stack it uses when changing exception level.

Saving the user stack is relevant when the scheduler swaps the context before the exception handler (here syscall handler) returns. I wasn't aware of the fact that qkernel can be preempted during the syscall handler when I submitted this PR, **I'll update this PR soon regarding saving user/kernel SP**

> Could you please give some explainations of why these registers are saved?

I'll elaborate more when I finish the changes above. But in general:
- x0 ~ x30 are saved, regardless the calling convention. Because the same routine also handles irq, which should be transparent to the control flow.
- tpidr_el0: for TLS, but this may deem unnecessary.
- esr_el1: Exception Syndrome Register, holds the "exception type", I think this is relevant when nested exceptions are allowed.
- elr_el1: Exception Link Register, holds the return address when taking an exception to EL1
- spsr_el1 : Saved Program Status Register, holds the PSTATE before the exception.

shrik3:
@CharlyYu please check the update. The sp is added to the exception frame. (hopefully this resolves the previous discussions)

qkernel/src/lib.rs
```diff
@@ -332,6 +330,113 @@ pub extern "C" fn syscall_handler(
...
+        } else {
+            nr = SysCallID::UnknowSyscall as _;
+            SysCallID::UnknowSyscall
+        };
+
+        if llevel == LogLevel::Complex {
+            tid = currTask.Thread().lock().id;
+            pid = currTask.Thread().ThreadGroup().ID();
+            info!("({}/{})------get call id {:?} arg0:{:x}, 1:{:x}, 2:{:x}, 3:{:x}, 4:{:x}, 5:{:x}, userstack:{:x}, return address:{:x}, fs:{:x}",
+                tid, pid, callId, _arg0, _arg1, _arg2, _arg3, _arg4, _arg5, currTask.GetPtRegs().rsp, currTask.GetPtRegs().rcx, GetFs());
```
CharlyYu, line 392: I think it may not compile if we define a PtRegs in aarch64, and I think maybe we have to union the definition of `ExceptionFrame` and `PtRegs`, because qkernel also get this PtRegs from the stack.  I made a temporary definition of `PtRegs` in aarch64 in this PR https://github.com/QuarkContainer/Quark/pull/1027, could you please take a look?

shrik3: regarding the `GetPtRegs`, I'm aware that this won't work for aarch64 right now. For now the `syscall_dispatch_aarch64` is a mere copy of the x86 syscall handler code (i.e. placeholder).

I'll get this fixed when we pin down the `ExceptionFrame` definitions. 

> could you please take a look?

Yes, and thanks for the ping. Cheers.

shrik3:
> Could you please add the pagefault handler?

Yes. I'll work on that.

shrik3:
> @shrik3 I have some code regarding ISS, and further debugging (as I am stuck in an EL1_SYNC_DataAbort-Fault). I can share the file with you if you wish...

@chl337 it would be nice if you could drop me a patch.

chl337:
> > @shrik3 I have some code regarding ISS, and further debugging (as I am stuck in an EL1_SYNC_DataAbort-Fault). I can share the file with you if you wish...
> 
> @chl337 it would be nice if you could drop me a patch.

(Mind this is experimental:)
- [add_iss_path_qkernel_src_interrupt_aarch64_mod_rs.diff](https://jmp.sh/LISqxrOv), if _git apply_ does not work just copy-past.
- [path_qkernel_src_interrupt_aarch64_fault_handler_rs](https://gist.github.com/rust-play/12f65601f0355428224c32de907e5845)

This should add some debugging information. With the real fault handler, I am not done yet.

shrik3:
> This should add some debugging information. With the real fault handler, I am not done yet.

@chl337 Thanks!  Also IMO I would suggest we put the extensive ESR helpers/defs in a separate module rather than the root mod.rs. Besides ISS for data abort, we have a bunch of defs for instr aborts as well and this would grow very fast ...

shrik3:
> > @chl337 Thanks! Also IMO I would suggest we put the extensive ESR helpers/defs in a separate module rather than the root mod.rs. Besides ISS for data abort, we have a bunch of defs for instr aborts as well and this would grow very fast ...
> 
> I think we should also implement instruction abort as soon because once we enter the el0, the first exception would be instruction abort as the PC points to an address that PTE is missing.

It's WIP, I'll push updates today or tomorrow.

shrik3:
On 23/12/05 07:35AM, CharlyYu wrote:
> Could you please add the pagefault handler? 
> 
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/QuarkContainer/Quark/pull/1009#issuecomment-1841037581
> You are receiving this because you were mentioned.
> 
> Message ID: ***@***.***>

regarding code structure, what's the proper place for the page fault handler
(like linux `mm/fault.c`)?

chl337:
> On 23/12/05 07:35AM, CharlyYu wrote: Could you please add the pagefault handler? -- Reply to this email directly or view it on GitHub: [#1009 (comment)](https://github.com/QuarkContainer/Quark/pull/1009#issuecomment-1841037581) You are receiving this because you were mentioned. Message ID: ***@***.***>
> regarding code structure, what's the proper place for the page fault handler (like linux `mm/fault.c`)?

@shrik3 This is exclusively a _qkernel_ thing, so it should be in _qkernel/src/.../_, e.g. _qkernel/src/interrupt/aarch64/mm/fault_handler.rs_.

shrik3:
On 23/12/07 07:54AM, chl337 wrote:
> @shrik3 This is exclusively a _qkernel_ thing, so it should be in _qkernel/src/.../_, e.g. _qkernel/src/interrupt/aarch64/mm/fault_handler.rs_.  


although I don't think /interrupt/{arch}/mm/fault_handler fits the semantics ..
(I mean why should mm fault handler be part of interrupt module?), for now I
would put a diagnostic data/instr abort handler in the.

/interrupt/aarch64/mod.rs , the openBSD code does some similar:

in `arch/arm64/arm64/trap.c` :

`udata_abort(struct trapframe *frame, uint64_t esr, uint64_t far, int exe) { ... }`

chl337:
> On 23/12/07 07:54AM, chl337 wrote: @shrik3 This is exclusively a _qkernel_ thing, so it should be in _qkernel/src/.../_, e.g. _qkernel/src/interrupt/aarch64/mm/fault_handler.rs_.
> although I don't think /interrupt/{arch}/mm/fault_handler fits the semantics .. (I mean why should mm fault handler be part of interrupt module?), for now I would put a diagnostic data/instr abort handler in the. /interrupt/aarch64/mod.rs , the openBSD code does some similar: in `arch/arm64/arm64/trap.c` : `udata_abort(struct trapframe *frame, uint64_t esr, uint64_t far, int exe) { ... }`

It meant it should be placed somewhere in _qkernel_, sorry for the confusion. Maybe it is time to introduce a _qkernel/src/arch/_ directory for future code (and refactor the old one).

> the openBSD code does some similar: in `arch/arm64/arm64/trap.c` 
Thanks for the code reference, as I needed some guidance :)

shrik3:
On 23/12/07 08:36AM, chl337 wrote:
> It meant it should be placed somewhere in _qkernel_, sorry for the confusion. Maybe it is time to introduce a _qkernel/src/arch/_ directory for future code (and refactor the old one).
> 

We have a stack of unmerged code atm, to avoid conflicts I would suggest we flip
the directory after #1009 / #1027 is done.

chl337:
@shrik3 Are you sure the values for _ESR_CM_ and _ESR_WNR_ correct?
Otherwise, please apply [this](https://jmp.sh/zrn5zIGb).

shrik3:
> @shrik3 Are you sure the values for _ESR_CM_ and _ESR_WNR_ correct? Otherwise, please apply [this](https://jmp.sh/zrn5zIGb).

Thanks, it's my fault, I meant bit shift, i.e. 1<<8 instead of 8. The PR is updated.

Also as for syntactic, I suggest that masks and values be written differently. i.e.

A = 0x100 is a value while
A = 0x1 << 8 is a mask.

qkernel/src/interrupt/aarch64/mod.rs
```diff
@@ -1,6 +1,309 @@
...
+pub extern "C" fn exception_handler_el0_fiq(){return;}
+#[no_mangle]
+pub extern "C" fn exception_handler_el0_serror(){return;}
+
+#[inline]
+pub fn GetFaultAccessType(esr:u64, is_exe:bool) -> AccessType{
+    if is_exe {
+        return AccessType(MmapProt::PROT_EXEC);
+    }
+    if !EsrDefs::IsCM(esr) && EsrDefs::IsWnR(esr){
```
chl337, line 272: @shrik3 I am not sure about the logic here. If `CM==1 -> IsCM==T -> !IsCM==F -> expression equals F.`
I would suggest `IsCM() || IsWnR()`.

CharlyYu:
Can we merge this pr? @chl337 @SchwarzStein

shrik3:
> Can we merge this pr? @chl337 @SchwarzStein

@SchwarzStein is on vacation.  

@CharlyYu You could merge if this looks fine to you.

chl337:
> Can we merge this pr? @chl337 @SchwarzStein

From my side, yes.

PR #1011, opened 2023-10-30
Fix context issue

CharlyYu:
1. fix allocator address issue
    2. make Context in Task a arch agnostic struct
    3. use ttbr0_el1 as the init sp, no kernel sp used
    4. fix pagetable fork issue, add map of mmio address
    5. support context swap

### DIFF
--- a/qlib/kernel/arch/aarch64/arch_def.rs
+++ b/qlib/kernel/arch/aarch64/arch_def.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use core::{slice, sync::atomic::{AtomicUsize, Ordering}};
+use core::{slice, sync::atomic::{AtomicUsize, AtomicU64, Ordering}};
 use alloc::sync::Arc;
 
 use crate::qlib as qlib;
@@ -193,5 +193,87 @@ impl State {
             Regs: regs,
             archFPState: Arc::new(QMutex::new(self.archFPState.lock().Fork()))
         }
+    }
+}
+
+pub type Context = Aarch64Context;
+
+#[derive(Debug)]
+#[repr(C)]
+pub struct Aarch64Context {
+    pub x19: u64,
+    pub x20: u64,
+    pub x21: u64,
+    pub x22: u64,
+    pub x23: u64,
+    pub x24: u64,
+    pub x25: u64,
+    pub x26: u64,
+    pub x27: u64,
+    pub x28: u64,
+    pub fp: u64,
+    pub sp: u64,
+    pub pc: u64,
+    pub x0: u64,
+    pub ready: AtomicU64,
+    pub tls: u64,
+}
+
+impl Aarch64Context {
+    pub fn New() -> Self {
+        return Self {
+         x19:    0,
+         x20:    0,
+         x21:    0,
+         x22:    0,
+         x23:    0,
+         x24:    0,
+         x25:    0,
+         x26:    0,
+         x27:    0,
+         x28:    0,
+         fp:     0,
+         sp:     0,
+         pc:     0,
+         x0:     0,
+         ready:  AtomicU64::new(1),
+         tls:    0,
+        };
+    }
+
+    pub fn set_tls(&mut self, tls: u64) {
+        self.tls = tls;
+    }
+
+    pub fn get_tls(&self) -> u64 {
+        self.tls
+    }
+
+    pub fn set_sp(&mut self, sp: u64) {
+        self.sp = sp;
+    }
+
+    pub fn get_sp(&self) -> u64 {
+        self.sp
+    }
+
+    pub fn set_para(&mut self, para: u64) {
+        self.x0 = para;
+    }
+
+    pub fn get_pc(&self) -> u64 {
+        self.pc
+    }
+
+    pub fn set_pc(&mut self, pc: u64) {
+        self.pc = pc;
+    }
+
+    pub fn set_ready(&self, val: u64) {
+        self.ready.store(val, Ordering::Release);
+    }
+
+    pub fn get_ready(&self) -> u64 {
+        return self.ready.load(Ordering::Acquire);
     }
 }
--- a/qkernel/src/kernel_def.rs
+++ b/qkernel/src/kernel_def.rs
@@ -217,7 +217,7 @@ pub fn switch(from: TaskId, to: TaskId) {
     if !SHARESPACE.config.read().KernelPagetable {
         toCtx.SwitchPageTable();
     }
-    toCtx.SetFS();
+    toCtx.SetTLS();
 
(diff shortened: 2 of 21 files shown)


qlib/kernel/arch/aarch64/arch_def.rs
```diff
@@ -195,3 +195,67 @@ impl State {
...
+	pub x23: u64,
+	pub x24: u64,
+	pub x25: u64,
+	pub x26: u64,
+	pub x27: u64,
+	pub x28: u64,
+	pub fp: u64,
+	pub sp: u64,
+	pub pc: u64,
+    pub tls: u64,
```
shrik3, line 217: seems like a mix of space/tab indentation

shrik3:
**clock arithmetic overflows** 

after some time of running (dozen seconds) this error message pops out:

```
[0/40004365] Unable to update params: Common("old now time computation overflowed2. params = Parameters { BaseCycles: 125214, BaseRef: 19777724432452382, Frequency: 1000000000 }, now = 1000108435").
```

The error is reported from 

```
qlib/kernel/kernel/timer/parameters.rs:149
```

chl337:
Hi @CharlyYu, regarding FP initialisation, could you try this [patch](https://jmp.sh/s/v1oiIgdn4ahnKql19MB4) and let me know of any change.
Could you also post the result of `lscpu | grep "Flags"`

CharlyYu:
Hi @chl337  this is the result on my raspberrypi
```
feng@raspberrypi:~$ lscpu | grep "Flags"
Flags:                              fp asimd evtstrm crc32 cpuid

```

CharlyYu:
> could you try this [patch](https://jmp.sh/s/v1oiIgdn4ahnKql19MB4) and let me know of any change.
The patch works

CharlyYu:
> clock arithmetic overflows

This must be a bug, and I will debug it later, but I think it is not related with this PR.

qlib/kernel/arch/aarch64/arch_def.rs
```diff
@@ -195,3 +195,67 @@ impl State {
         }
     }
 }
+
+pub type Context = Aarch64Context;
+
+#[derive(Debug)]
+#[repr(C)]
+pub struct Aarch64Context {
```
SchwarzStein, line 203: is this the context to use for all situations ?

CharlyYu: I'm not quite sure, @chl337 is taking care of the context swap thing, he can add or delete registers from this struct if it is necessary.

chl337: As I understand it, this structure is used for task switching, e.g. after a scheduling decision... That said, AFAIK at least the _calle-saved, FP, LR, SP_ registers must be preserved, **ref:** [ABI](https://github.com/ARM-software/abi-aa/blob/2a70c42d62e9c3eb5887fa50b71257f20daca6f9/aapcs64/aapcs64.rst#general-purpose-registers), [Context Switch](https://developer.arm.com/documentation/den0024/a/The-Memory-Management-Unit/Context-switching).

chl337:
Hi @CharlyYu, could you please check if this [patch](https://jmp.sh/s/V0N3LZExWyuA27sxo5vQ) works for you exactly the same as the status of this PR. It is meant to resolve possible future merge conflicts between us, as it seems we were touching some similar places. Please let me know in case of any error or preferences!
P.S. In case of errors, I can look at them first from Monday on. Sorry for the inconvenience!

qlib/kernel/arch/aarch64/arch_def.rs
CharlyYu: Actually FP LR SP are included in the struct, only the LR is actually save in the field of name `pc`

SchwarzStein: wouldn't you want to save all the context x0-x31, especially when you do a conditional yield.
e.g.
```
 while( !check_asynchronous_event() ) {
     yield();
}
```

SchwarzStein:
@CharlyYu , have you tested the refactored code in x86? Do we still run with no issues?

CharlyYu:
> @CharlyYu , have you tested the refactored code in x86? Do we still run with no issues?

not yet

SchwarzStein:
> > @CharlyYu , have you tested the refactored code in x86? Do we still run with no issues?
> 
> not yet

Any suggestions on how we proceed. 
Should the above be sufficient and I merge this or there is more you want to add to this?

qlib/kernel/arch/aarch64/arch_def.rs
CharlyYu: I think not all the registers saved in X86 context too, I am not sure how the `yield` is supported, maybe it is supported by some signal mechanism.

CharlyYu:
I am still working on the context swap thing, but currently it is blocked in the `LoadUserTable`. even I changed the codes to
```
#[inline]
pub fn LoadUserTable(table: u64) {
    unsafe {
        asm!("msr ttbr0_el1, {0}
        ic iallu
        dsb nsh
        isb
        ", 
        in(reg) table);
    };
}
```

The process is still blocked in this function call. I don't get the reason, maybe you guys can find it if it is convenient.

CharlyYu:
> Should the above be sufficient and I merge this or there is more you want to add to this?

I have some code changes not submited. maybe we can merge this PR after the `LoadUserTable` thing get fixed.

SchwarzStein:
> > Should the above be sufficient and I merge this or there is more you want to add to this?
> 
> I have some code changes not submited. maybe we can merge this PR after the `LoadUserTable` thing get fixed.

Could you make an rfc, and I will have a look at the entire changes for what might be causing this.
If you can point out the code you use to create the tables and configure the registers, I might spot the bug faster.

CharlyYu:
@SchwarzStein I have got the reason, it is the forked pagetable did not map the mmio address.

CharlyYu:
> Hi @CharlyYu, could you please check if this [patch](https://jmp.sh/s/V0N3LZExWyuA27sxo5vQ) works for you exactly the same as the status of this PR

@chl337 The context swap seems ok now with codes change in this PR now, I have implemented a preliminary version of `context_swap` assembly function in the `qlib/kernel/arch/aarch64/syscall_entey.s`. Could you please take a look at code changes in this PR? and there is any error or  omission in this imlemetation, it is welcome to leave you comment, or you can submit a new PR to correct them.

CharlyYu:
> Should the above be sufficient and I merge this or there is more you want to add to this?

@SchwarzStein I think this PR is ready to be merged after clh337's approve. Please take a look, @chl337

CharlyYu:
@chl337 Could you please take a special look at the memory barries instructions in the `context_swap`?  I am not sure if it is appropriate to do it like that.

PR #1013, opened 2023-11-06
aarch64: fix slice copy issue

CharlyYu:
the length of source and destination slice should be equal otherwise it will panic to call `copy_from_slice`

### DIFF
--- a/qvisor/src/ucall/usocket.rs
+++ b/qvisor/src/ucall/usocket.rs
@@ -59,12 +59,7 @@ impl USocket {
 
         let cstr = CString::New(path);
         let slice = cstr.Slice();
-        #[cfg(target_arch = "x86_64")]
-        for i in 0..slice.len() {
-            server.sun_path[i] = slice[i] as i8;
-        }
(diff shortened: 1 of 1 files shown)


chl337:
As this is cause by `copy_from_slice`, maybe there is no need for two "architectural" separate ways to do it. Maybe apply this [unified change](https://jmp.sh/exF5xSuR)

CharlyYu:
You are right, I just didn't want to change the logic of x86_64,but it is better change it in a unified way.

CharlyYu:
@chl337  done

CharlyYu:
@chl337 If it is ok for you, I will merge this

chl337:
> @chl337 If it is ok for you, I will merge this

Yes, if needed can be adjusted later.

CharlyYu:
> As this is caused by copy_from_slice, maybe there is no need for two "architectural" separate ways to do it. Maybe apply this [unified change](https://jmp.sh/exF5xSuR)

@chl337  I already make this unified.

PR #1014, opened 2023-11-06
Armdev - Removed non-compatible flag from mmap(...) in QVisor 

chl337:
In "qvisor/(SOME_LONG_PATH)/runtime/vm.rs", the _MAP_FIXED_ flag in ```mmap(addr, ..., MAP_FIXED ...)``` causes _mmap_ to take the _addr_ parameter as fact, which in this call is ```std::ptr::null_mut()```, leading in undesired behavior. The mapping is not related to an opened file, _fd = -1_.

Late addition, fixed initialization of CPACR_EL1 register, related to FP-functionality.


CharlyYu:
LGTM

PR #1015, opened 2023-11-07
aarch64：distinguish between x86 and aarch64 for the definition of LibcStat

CharlyYu:
The struct definition of `stat` is slightly different in x86 and aarch64

### DIFF
--- a/qlib/linux_def.rs
+++ b/qlib/linux_def.rs
@@ -3243,6 +3243,7 @@ pub fn CopyPage(to: u64, from: u64) {
     }
 }
 
+#[cfg(target_arch = "x86_64")]
 #[derive(Debug, Default, Copy, Clone)]
 #[repr(C)]
 pub struct LibcStat {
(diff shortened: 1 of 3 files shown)


qlib/linux_def.rs
```diff
@@ -3266,6 +3267,31 @@ pub struct LibcStat {
     pub pad: [i64; 3],
 }
 
+#[cfg(target_arch = "aarch64")]
+#[derive(Debug, Default, Copy, Clone)]
+#[repr(C)]
+pub struct LibcStat {
```
chl337, line 3273: I can only attest that _struct LibcStat_ matches this [cross-reference](https://chromium.googlesource.com/android_tools/+/20ee6d20/ndk/platforms/android-21/arch-arm64/usr/include/sys/stat.h#40). Probably @SchwarzStein and @shrik3 have more insight.

shrik3: it's maybe better to check against the [linux header defs](https://elixir.bootlin.com/linux/v6.6/source/arch/arm64/include/asm/stat.h#L26) for this.

CharlyYu: Actually, I followed the definition in the libc crate in rust  https://github.com/rust-lang/libc/blob/main/src/unix/linux_like/linux/gnu/b64/aarch64/mod.rs#L53 

> it's maybe better to check against the [linux header defs](https://elixir.bootlin.com/linux/v6.6/source/arch/arm64/include/asm/stat.h#L26) for this.

It seems the definition in the link is for the compat and it is quite different from the 64 bits. @shrik3

chl337: JFI, the reference to [generic `struct stat` in linux](https://elixir.bootlin.com/linux/v6.6/source/include/uapi/asm-generic/stat.h#L24). If my eyes don't lie, it matches the one from [Rust](https://github.com/rust-lang/libc/blob/main/src/unix/linux_like/linux/gnu/b64/aarch64/mod.rs#L53) that @CharlyYu refers to.

CharlyYu: @shrik3 If it is ok for you I will merge this pr

shrik3: > @shrik3 If it is ok for you I will merge this pr

I'm ok with this.

PR #1018, opened 2023-11-13
aarch64: add functions to set/clear PAN

shrik3:
add functions to set or clear the PAN bit in PSTATE.

This won't be needed until a proper copy_{to,from}_user is added.

### DIFF
--- a/qlib/kernel/asm/aarch64.rs
+++ b/qlib/kernel/asm/aarch64.rs
@@ -639,3 +639,22 @@ pub fn spsel() -> u64 {
     }
     ret
 }
+
+// Setting or clearing the (hardware) privilege access never bit. If set, the
+// CPU will be trapped when trying to access EL0(user) memory from EL1 or above.
+// This is to prevent the kernel from accessing user memory accidentally,
+#[cfg(target_arch = "aarch64")]
+#[inline]
+pub fn SetPAN() {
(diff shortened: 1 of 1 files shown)


qlib/kernel/asm/aarch64.rs
```diff
@@ -639,3 +639,28 @@ pub fn spsel() -> u64 {
...
+
+// Setting or clearing the (hardware) privilege access never bit. If set, the
+// CPU will be trapped when trying to access EL0(user) memory from EL1 or above.
+// This is to prevent the kernel from accessing user memory accidentally,
+#[cfg(target_arch = "aarch64")]
+#[inline]
+pub fn SetPAN() {
+    unsafe {
+        asm!(
+            "msr PAN, #1",
```
CharlyYu, line 651: why does ClearPAN need the memory barrier, and SetPAN doesn't?

shrik3: on a second thought the fences should not be here, sorry.

I'll delete

shrik3:
I'm not sure if `MSR PAN, #1` with immediate value really works, but this is in the ARM manual.

shrik3:
This will be reworked in another PR, where the PAN functions are wrapped with target_feature conditions.

PR #1024, opened 2023-11-24
Fix for Docker start behavior with systemctl

cxyzhao:
- Identified an issue where using 'sudo systemctl restart docker' does not always forcibly restart the 'docker.service'.
- Implemented a check to ensure that 'sudo systemctl restart docker.service' is explicitly called for a consistent restart behavior.
- This fix aims to address the inconsistent behavior observed in some cases when restarting Docker using systemctl commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ sudo systemctl restart docker
 The helloworld docker sample application can be executed as below:
 ```sh
 sudo systemctl restart docker
+sudo systemctl restart docker.service
 docker run --rm --runtime=quark hello-world
 ```      
     


shrik3:
It may not be a good idea to comment on a closed PR but out of curiosity:

isn't `systemctl command docker` supposed to be identical to 
`systemctl command docker.service` ? 

man systemctl(1)

```
[...] If the suffix is not specified (unit name is "abbreviated"), systemctl
will append a suitable suffix, ".service" by default, and a type-specific 
suffix in case ofcommands which operate only on specific unit types. 

For example,
    # systemctl start sshd
and
    # systemctl start sshd.service
are equivalent [...]
```

I think running the systemctl command with docker and docker.service would most likely end up restarting docker twice.

PR #1027, opened 2023-11-30
Support enter user for aarch64

CharlyYu:


CharlyYu:
Something is wrong, it will be an exception of number 32 when we get into the user state.
Could you please take a look at this? @shrik3 @chl337

qlib/kernel/asm/aarch64.rs
```diff
@@ -104,9 +131,32 @@ pub fn SyscallRet(kernelRsp: u64) -> ! {
 #[inline]
 pub fn IRet(kernelRsp: u64) -> ! {
     unsafe {
-        asm!(
-            "nop"
-        );
+        asm!("\
```
shrik3, line 134: Maybe we should keep in sync wrt. the frame format?

https://github.com/QuarkContainer/Quark/blob/9d25d239b506d45ac804d9915d2fffafbe4f1703/qlib/kernel/arch/aarch64/exception.s#L17

chl337:
@CharlyYu I will look at it.

shrik3:
@CharlyYu  a meta question, this PR is based against armdev, and doesn't include the #1009 right?

But the error message comes from the code in #1009.  Should I merge this PR to #1009 to reproduce the error you have?

chl337:
@CharlyYu can you please describe how to reproduce the error?

chl337:
> @CharlyYu a meta question, this PR is based against armdev, and doesn't include the #1009 right?
> 
> But the error message comes from the code in #1009. Should I merge this PR to #1009 to reproduce the error you have?

@CharlyYu also please mention which version of the above-mentioned PR.

CharlyYu:
> But the error message comes from the code in https://github.com/QuarkContainer/Quark/pull/1009. Should I merge this PR to https://github.com/QuarkContainer/Quark/pull/1009 to reproduce the error you have?

Yes I did the test based on the merge of  #1009 #1013 #1015

CharlyYu:
> @CharlyYu can you please describe how to reproduce the error?
And I did the test by run `docker run -d --runtime=quark_d nginx`

chl337:
@CharlyYu Thank you for the information.

qlib/kernel/asm/aarch64.rs
CharlyYu: Yes, That what I mentioned in this comment https://github.com/QuarkContainer/Quark/pull/1009#discussion_r1410957231, I think we should first merge the definition of `PtRegs` and `ExceptionFrame`

chl337:
> Something is wrong, it will be an exception of number 32 when we get into the user state. Could you please take a look at this? @shrik3 @chl337 

@CharlyYu I am afraid, I cannot reproduce the same problem. Instead, I get a SYNC_exception from EL1. Maybe @shrik3 has more insight.
```
b/qkernel/src/interrupt/aarch64/mod.rs

#[no_mangle]
pub extern "C" fn exception_handler_unhandled(_exception_frame_ptr:usize, exception_type: u8){
    panic!("unhandled exception - {}", match(exception_type) {
        1 => "EL1_SYN",
                ....
        _ => "NON-DEFINED",

    } 
        );
}

#[no_mangle]
pub extern "C" fn exception_handler_el1h_sync(exception_frame_ptr:usize){
    return exception_handler_unhandled(exception_frame_ptr, 1);
}

#[no_mangle]
pub extern "C" fn exception_handler_el1h_irq(exception_frame_ptr:usize){
    return exception_handler_unhandled(exception_frame_ptr, 2);
}
            ..........
#[no_mangle]
pub extern "C" fn exception_handler_el0_sync(exception_frame_ptr:usize){
            ...............
        _ => {
            panic!("unhandled sync exception from el0: {}\n", ec);
        }
    }
}
```

CharlyYu:
@chl337  It maybe a pagefault exception when executing the instruction but the file is not actually loaded into the memory yet. and the vma is mapped to the binary file. Could you please add the handler of pagefault exception? @shrik3

chl337:
> @chl337 It maybe a pagefault exception when executing the instruction but the file is not actually loaded into the memory yet. and the vma is mapped to the binary file. Could you please add the handler of pagefault exception? @shrik3

@CharlyYu Indeed, maybe that is the case... I will try to implement a dummy handler to verify it.

shrik3:
@CharlyYu @chl337 Please take a look at the updated commit

Now we are looking at a translation fault at EL1. To add proper page fault handler I suggest we  merge the opening PRs first to avoid VCS pasta.

Reprod:
```
git checkout armdev
# git merge #1009, #1013, #1015 and #1027 in order
# set debug level
make debug  && make install
docker run -d --rm --runtime=quark_d hello-world
```

Log:
```
[DEBUG] [1/43c0140000|0] get data abort fault from el1
[Print] [1/43c0140000|0] get panic : Some(DFSC/IFSC == 0x05, FAR == 4198388)
[Print] [1/43c0140000|0] panic occurred in file 'src/interrupt/aarch64/mod.rs' at line 306
```

qlib/kernel/SignalDef.rs
```diff
@@ -89,6 +87,45 @@ impl PtRegs {
...
+    }
+
+    pub fn set_stack_pointer(&mut self, sp: u64) {
+        self.rsp = sp;
+    }
+}
+
+#[cfg(target_arch = "aarch64")]
+#[repr(C)]
+#[derive(Debug, Copy, Clone, Default)]
```
shrik3, line 102: Could you add a custom Debug trait for aarch64 SigContext?

The __reserved field doesn't not need to be printed, and it is too large, making the log very noisy. (4k lines printed per run).  The following snippet would work:

```rust
impl fmt::Debug for SigContext {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.debug_struct("SigContext")
            .field("fault_address", &self.fault_address)
            .field("Regs", &self.regs)
            .field("sp", &self.sp)
            .field("pc", &self.pc)
            .field("pstate", &self.pstate)
            .finish()
    }
}

```

chl337: @shrik3 I can create a patch for this.

chl337:
> @CharlyYu @chl337 Please take a look at the updated commit in #1009
> 
> Now we are looking at a translation fault at EL1. To add proper page fault handler I suggest we merge the opening PRs first to avoid VCS pasta.
> 
> Reprod:
> 
> ```
> git checkout armdev
> # git merge #1009, #1013, #1015 and #1027 in order
> # set debug level
> make debug  && make install
> docker run -d --rm --runtime=quark_d hello-world
> ```
> 
> Log:
> 
> ```
> [DEBUG] [1/43c0140000|0] get data abort fault from el1
> [Print] [1/43c0140000|0] get panic : Some(DFSC/IFSC == 0x05, FAR == 4198388)
> [Print] [1/43c0140000|0] panic occurred in file 'src/interrupt/aarch64/mod.rs' at line 306
> ```
> 
> [edit] : with CopyDataWithPF disabled, I got the same Error: here is the diagnostics:
> 
> ```
> [DEBUG] [1/43c0140000|0] get instruction abort fault from el0
> [Print] [1/43c0140000|0] get panic : Some(DFSC/IFSC == 0x07, FAR == 4194680, faulty access type = --x)
> ```
> 
> That suggests a Translation fault at LV3, I think it's a missing or misconfigured PTE? Since the data is actually copied out to user, there shouldn't be a page fault.

I have a different situation:
> load root process failure with error SysError(12), shutting down...

CharlyYu:
> Please review 1009 and, if there is no further problem, I propose we merge the still-opening PRs first.

#1009  is merged, looking forward the pr porting PF handler. @chl337

chl337:
> > Please review 1009 and, if there is no further problem, I propose we merge the still-opening PRs first.
> 
> #1009 is merged, looking forward the pr porting PF handler. @chl337

@CharlyYu Thanks. There is still some work to be done concerning page tables. I will open (later) a new WIP for the PF-handler.

qlib/kernel/SignalDef.rs
CharlyYu: done

PR #1031, opened 2023-12-11
[RFC]VMM: adjust number of vCPUS for VM with config.json

chl337:
Enable configuring the amount of VM's vCPUs from Quark's configuration file. This could be useful especially during development when we want minimal setups.

### DIFF
--- a/config.json
+++ b/config.json
@@ -27,5 +27,6 @@
   "TlbShootdownWait": true,
   "Sandboxed"     : false,
   "Realtime"      : false,
-  "EnableIOBuf"   : false
+  "EnableIOBuf"   : false,
+  "LimitVMvCPUAmount": 0
 }
--- a/qlib/config.rs
+++ b/qlib/config.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::qlib::MIN_VCPU_COUNT;
+
 #[derive(Clone, Copy, Debug, Serialize, Deserialize)]
 pub struct Config {
     pub DebugLevel: DebugLevel,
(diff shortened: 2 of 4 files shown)


config.json
```diff
@@ -27,5 +27,6 @@
   "TlbShootdownWait": true,
   "Sandboxed"     : false,
   "Realtime"      : false,
-  "EnableIOBuf"   : false
+  "EnableIOBuf"   : false,
+  "SuggestedVMvCPUAmount": 0
```
shrik3, line 31: IHO "Max..." , or "...Limit"  would be more sensible naming than "Suggested".

qlib/config.rs
```diff
@@ -54,6 +57,19 @@ impl Config {
...
         return self.LogType == LogType::Async;
     }
+
+    //
+    // The assigned value is taken in consideration
+    // if it is at least equal to the required
+    // MIN_VCPU_COUNT value.
+    //
+    pub fn get_config_vcpu_amount(&self) -> Option<usize> {
+       if self.LimitVMvCPUAmount < MIN_VCPU_COUNT as u64 {
```
CharlyYu, line 67: Why is there such a limitation that the minimal vcpu count should be greater than 2 ?

CharlyYu (commented):
Actually we need to adjust the vcpu count of vm according to the container spec, but it is not implemented yet

qlib/config.rs
chl337: It sets the limit to at least two because vCPU_0 is used for [IOWait();](https://github.com/QuarkContainer/Quark/blob/483b83c1f51958f0b1cbc67667a2a311bfe39647/qkernel/src/lib.rs#L538)

shrik3:
FOr testing purposes, you could simply pass -cpus=<n> parameter to the `docker run` cli to limit the number of vcpus.

chl337:
On Mon May 27, 2024 at 11:53 AM CEST, Tianhao Wang wrote:
> FOr testing purposes, you could simply pass -cpus=<n> parameter to the `docker run` cli to limit the number of vcpus.
Thanks, didn't know that!

PR #1034, opened 2023-12-18
armdev - PageFault handler

chl337:
**Important: this PR is undergoing rebasing**; updates through force push happen; please be careful if you work on this branch locally.

WIP on page fault handler for the armdev-brach.

~~**NOTE:** Required patches for integration with [PR1027](https://github.com/QuarkContainer/Quark/pull/1027) will be provided shortly.~~
~~**EDIT:** Commits 561c1fc, f7d3c9c and 40f108b from [PR1027](https://github.com/QuarkContainer/Quark/pull/1027) are included, to facilitate testing.~~


qkernel/src/interrupt/aarch64/mod.rs
```diff
@@ -285,9 +299,28 @@ pub fn MemAbortUser(ptregs_addr:usize, esr:u64, far:u64, is_instr:bool){
     let dfsc = esr & EsrDefs::ISS_DFSC_MASK;
     let access_type = GetFaultAccessType(esr, is_instr);
     match dfsc {
+        EsrDefs::DFSC_TF_L3 => {
```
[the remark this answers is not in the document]
chl337: I will replace it accordingly.

qlib/linux_def.rs
```diff
@@ -3266,6 +3267,31 @@ pub struct LibcStat {
     pub pad: [i64; 3],
 }
 
+#[cfg(target_arch = "aarch64")]
+#[derive(Debug, Default, Copy, Clone)]
+#[repr(C)]
+pub struct LibcStat {
```
shrik3, line 3273: isn't this the same as done in https://github.com/QuarkContainer/Quark/pull/1015/commits/e1469ade75fb7ae331e7d2ef562a2f7ef0a4dd79 ?

CharlyYu:
It seems the MapPage function does not work, it page fault into kernel with the same FAR repeatly

chl337:
> It seems the MapPage function does not work, it page fault into kernel with the same FAR repeatly

We are currently working on it.

shrik3:
> > It seems the MapPage function does not work, it page fault into kernel with the same FAR repeatly
> 
> We are currently working on it.

plus I had a stupid mistake in the exception table where I used the old reg save/restore routines instead of the new ptregs routine. (I'll fix it in the armdev very quick)

Anyways, with that fixed, I'm having endless PF exceptions ...

CharlyYu:
I'm not sure if the tlbi instruction is correct after we install page in pagetable, maybe it is because we do not correctly flush the tlb?

chl337:
> I'm not sure if the tlbi instruction is correct after we install page in pagetable, maybe it is because we do not correctly flush the tlb?

I am done inspecting the configuration of the registers in [VMM](https://github.com/QuarkContainer/Quark/blob/e5cf5186a43c2d79800150f326c759cc1cb4bbdd/qvisor/src/kvm_vcpu_aarch64.rs#L248) and TLB related code. It does not seem to be the case that there is something wrong with them. On the other side, after setting the _PAGE_-bit on the 3rd Level PTEs no _MMU fault_ happens (see ARM Arch. Ref. Man. ARMv8-A "D5.6.1 Types of MMu faults"). I suspect that something goes wrong during the root process loading, I will focus on that for now...

qlib/kernel/SignalDef.rs
```diff
@@ -109,6 +109,18 @@ pub struct PtRegs {
     pub __pad: u64,
 }
 
+#[cfg(target_arch = "aarch64")]
```
shrik3, line 112: Please drop this from the PR, as is already added in #1027

qlib/pagetable.rs
```diff
@@ -41,17 +71,50 @@ cfg_aarch64! {
                                                          PageTableEntry, PageTableIndex,
                                                          PageTableFlags};
 
+   //
+   // A swapt-out page has the bit-flag set.
+   //
+   fn is_pte_swapped(flags: PageTableFlags) -> bool {
```
shrik3, line 77: please use the functions from bitflags! e.g. .contains(), .insert(), .remove().

The primitive bitwise ops are error prone

shrik3:
Please do a rust-fmt on the newly added files

qlib/kernel/SignalDef.rs
chl337: @shrik3 Done.

chl337: ~> @shrik3 Done.~
Readded commits, as it was not covert on #1027.

qlib/pagetable.rs
chl337: Done.

abel-von (approved):

PR #1038, opened 2024-01-13
aarch64: add CopyPageUnsafe asm codes

CharlyYu:
CopyPageUnsafe is the asm codes to copy a page, without it, executables can not even been loaded.


qlib/kernel/arch/aarch64/syscall_entry.s
```diff
@@ -7,6 +7,52 @@
...
+	ldp	x10, x11, [x1, #64]
+	ldp	x12, x13, [x1, #80]
+	ldp	x14, x15, [x1, #96]
+	ldp	x16, x17, [x1, #112]
+
+	add	x0, x0, #256
+	add	x1, x1, #128
+1:
+    // PAGE_SIZE
+	tst	x0, #(0x1000 - 1)
```
shrik3, line 23: if the supplied address is not 4k aligned, this loop will not terminate properly (perhaps it would end at all).  I would suggest either 1) explicitly check for alignment in the wrapper function or 2) use `cmp` instead of `tst` for robustness.

Also I think the loop is unrolled too much? Perhaps 64 bytes in one iteration is enough.

Otherwise LGTM.

chl337: I agree with @shrik3; the alignment test should be done beforehand.

shrik3:
update: with the updated PF handler (#1034)  doesn't seem to work well with CopyPageUnsafe:

with the copypage added, I'm having translation fault from EL0. The faulting address is from the upper half, which is illegal.

Without this CopyPage everything works (docker run hello-world). This falls into the "it works, but why?" category :( . I mean this CopyPage impl is definitely needed...

chl337:
> update: with the updated PF handler (#1034) doesn't seem to work well with CopyPageUnsafe:
> 
> with the copypage added, I'm having translation fault from EL0. The faulting address is from the upper half (in my case `0x74696e692e006595`), which is illegal.
> 
> Without this CopyPage everything works (docker run hello-world). This falls into the "it works, but why?" category :( . I mean this CopyPage impl is definitely needed...

I will do some testing around it... But my first question would be, @CharlyYu why is this function not implemented in plain rust?

shrik3:
> I will do some testing around it... But my first question would be, @CharlyYu why is this function not implemented in plain rust?

That aside, I tried rust's stock ptr::copy_nonoverlapping, and the result is the same : an upper addr from EL0.

chl337:
> > I will do some testing around it... But my first question would be, @CharlyYu why is this function not implemented in plain rust?
> 
> That aside, I tried rust's stock ptr::copy_nonoverlapping, and the result is the same : an upper addr from EL0.

1) If the same result can be achieved with plain Rust, let's keep it so.
2) Do you have any clue, from where is _CopyPage(...)_ been called?

qlib/kernel/arch/aarch64/syscall_entry.s
CharlyYu: Do we still have to do the alignment? It seems all the callers of the CopyPageUnsafe did made sure the source is page alignment.

shrik3: > It seems all the callers of the CopyPageUnsafe did made sure the source is page alignment.

I'm fine with this, as long as we keep it in mind. A code comment would be nice, e.g.:

```
/// UNSAFE CODE! the caller must make sure the to and from address are 4k aligned.
```

abel-von: done

CharlyYu: done

PR #1058, opened 2024-01-28
armdev: cleanup page mapping

shrik3:
1. no longer map MMIO mem regions in the host side
2. clean up the definition of PageTableFlags, improve bitops syntax
3. ~more verbose debug info during PF~
4. ~add function to dump all valid pagetables~

**Note on merge yet** : will be rebased after #1034

### DIFF
--- a/qlib/addr.rs
+++ b/qlib/addr.rs
@@ -62,14 +62,14 @@ impl AccessType {
 
     #[cfg(target_arch = "aarch64")]
     pub fn NewFromPageFlags(flags: PageTableFlags) -> Self {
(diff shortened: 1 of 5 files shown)


abel-von:
So the mmap of MMIO region is no use at first?

shrik3:
> So the mmap of MMIO region is no use at first?

Yes afaik, if the guest physical memory is not backed by host virtual mapping, access to that memory will cause MMIO.  it doesn't hurt to keep the mapping. I'm removing it as a cleanup.

I'm using this patch and MMIO based hypercalls works fine.

CharlyYu (approved):

Issue #1071, opened 2024-02-05
Allow accessing MIDR_EL1 from userspace


shrik3:
After returning from sys_brk, program crashes with an exception class 0 from EL0, the issue could be glibc accessing MIDR_EL1. This is not allowed in our current setup.

The issue could perhaps be fixed if we allow accessing MIDR_EL1 from userspace.  Reference :  
https://lore.kernel.org/lkml/88bdcca8-4df4-15ff-2e88-c53255b1fe30@daynix.com/

we are digging that. @CharlyYu

shrik3:
the linux kernel emulates MRS instructions for EL0 by inserting code to the "unknown" exception handler. We can do the same.

Proposed:   
1. on a "unknown" el0 sync exception, dereference user PtRegs.pc to go get the faulting opcode.
2. decode the instruction and check if it's a MRS on MIDR_EL1 (optionally also expand this to others*)
3. read corresponding register and put the result back into PtRegs.<Xt>

* : see https://docs.kernel.org/arch/arm64/cpu-feature-registers.html#list-of-registers-with-visible-features

PR #1076, opened 2024-02-07
[WIP][EXPERIMENTAL]armdev: emulate MRS instruction on system-regs for EL0 

chl337:
This is (only) a proof concept regarding #1071.
NOTE: For testing is needed - #1038
REF: ARM ® Architecture Reference Manual ARMv8-A architecture profile

P.S: The [bad64-crate](https://docs.rs/bad64/latest/bad64/index.html) seems like a good option for instruction translation.


shrik3:
full implementation in #1077

PR #1087, opened 2024-02-12
(temp) Armdev mprotect workaround 

shrik3:
this is a temporary workaround for a mprotect  issue. The `traverse` function takes a lambda to modify all PTE flags in a certain virt addr range.  On aarch64, this overwrites the `PAGE` bit in the PTE, which breaks later paging.

This is not a clean fix. The PF handler will later add a catch-all path that fixes such conditions.


chl337:
dd47fb66fe4ab8543516ce6cead489288af4c584 should handle the issue.

PR #1092, opened 2024-02-15
aarch64: save tls when leaving el0

shrik3:
Save tpidr_el0 to task.context.tls when leaving EL0.

Fixes #1091

PR #1094, opened 2024-02-16
aarch64: remove x86 specific code

shrik3:
Also:
- guard unimplemented with todo!() macro
- remove SysArchPrctl for aarch64

### DIFF
--- a/qlib/kernel/asm/aarch64.rs
+++ b/qlib/kernel/asm/aarch64.rs
@@ -44,20 +44,6 @@ pub const KernelFlagsSet:u64 = _PSR_MODE_EL1h | _PSR_D_BIT | _PSR_A_BIT | _PSR_I
 // UserFlagsSet are always set in userspace.
 pub const UserFlagsSet:u64 = _PSR_MODE_EL0t;
 
-#[inline]
-pub fn WriteMsr(msr: u32, value: u64) {
-
-}
-
-#[inline]
-pub fn ReadMsr(msr: u32) -> u64 {
-    0
-}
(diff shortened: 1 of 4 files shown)


chl337:
@shrik3 Interestingly, I am doing the same thing! I suspect some of the strange behavior I am facing relates to some call to an empty implementation.

qlib/kernel/asm/aarch64.rs
```diff
@@ -186,6 +173,7 @@ pub fn GetCurrentUserSp() -> u64 {
 
 #[inline]
 pub fn Clflush(addr: u64) {
```
chl337, line 175: @shrik3 This is obsolete, as it is not used anywhere.

qlib/kernel/asm/aarch64.rs
```diff
@@ -72,6 +58,7 @@ pub fn GetVcpuId() -> usize {
 
 #[inline]
 pub fn Hlt() {
```
chl337, line 60: @shrik3 This method is obsolete, as it is not used anywhere.

shrik3:
> @shrik3 Interestingly, I am doing the same thing! I suspect some of the strange behavior I am facing relates to some call to an empty implementation.

afaik "calling an empty implementation" would only apply to 
1. FPU init
2. getting FS/GS.   Also the tls should not be causing issues ofter #1092 

Anyways I don't think these would cause problem. (are there other such empty implementation that I missed?)

abel-von (approved):

PR #1118, opened 2024-02-23
aarch64: include HYPERCALL_MMIO in Kernel memory-space.

chl337:
In aarch64, we use an extra page below the _KVM_IOEVENTFD_BASEADDR_ page for Hypercalls, _HYPERCALL_MMIO_BASE_. Till now, we have missed including it in the memory management bookkeeping.


CharlyYu (approved):

shrik3:
note that the issue persists if the config is set to CopyDataWithPf : false;

Issue #1109, opened 2024-02-19
is it possible to opt out cuda during compilation?


shrik3:
is it possible to opt out cuda dependency? qvisor on the current main branch will not compile where cuda dependency could not be satisfied.

```
  = note: /usr/bin/ld: cannot find -lcuda: No such file or directory
```

QuarkContainer:
@shrik3 So far I didn't find good way to opt out cuda so far. Would you please have a try with following commands to see whether it could workaround the compiler issue?

sudo apt-get install libelf-dev nvidia-driver-535

shrik3:
> @shrik3 So far I didn't find good way to opt out cuda so far. Would you please have a try with following commands to see whether it could workaround the compiler issue?
> 
> sudo apt-get install libelf-dev nvidia-driver-535

thanks for reply.
With cuda dependencies installed, quark builds and runs without issue.  Still, it would be nice if cuda could be optional.

Also it feels like a hassle to install nvidia drivers on a machine that has no nvidia hardware.. :smirk:

changhong94:
Fixed by pr 1164.

Issue #1105, opened 2024-02-19
armdev: x86_64 "leftovers" in the log file.


chl337:
In the log file, there are traces of _x86_64_ notions that are not related to _aarch64_. They could be leftovers of some previous c/p with no real effect but should be removed as they cause confusion.

chl337:
> Perhaps replace them with some arm "relevant" stuffs like
> 
> SP_EL1, VBAR, TPIDR, TTBR ... (if they are not yet printed).

Yes, that is a good suggestion.

PR #1132, opened 2024-03-04
[Draft] Add interface for arch-specific vCPU

chl337:
The finished work should solve #1105 and allow smoother integration of other architectures in the future.

NOTE: #1109 causes the build to fail.


qvisor/src/arch/x86_64/mod.rs
```diff
@@ -12,4 +12,166 @@
...
+use crate::arch::vCPU;
+use crate::runc::runtime::vm::{Dump, ClearDump};
+use crate::qlib::common::{KERNEL_FLAGS_SET, Error};
+use crate::qlib::linux_def::MemoryDef;
+
+pub mod context;
+pub mod vm;
+
+#[derive(Debug)]
+pub struct x86_64vCPU {
```
shrik3, line 33: The naming of x86_64vCPU struct looks kinda uncanny to me .. is the cargo linter ok with this?

chl337: > is the cargo linter ok with this?
Specific on this, yes, but in general, no ;)
>naming looks uncanny to me
Happy to change it with any suggestion.

shrik3:
I don't have a better idea, that's the nature of this thingy. If linter is ok
then I'm ok :)

shrik3:
we really need this. (as soon as we merge with main)

`#[cfg(target_arch="xyz")]` here and there (inline) is annoying and error prone. Luckily we have only two architectures here. Imagine having more.

shrik3:
is this on the same agenda of #1136 , as well as the ongoing plan to abstract VM/VCPU creation?

PR #1133, opened 2024-03-04
aarch64: do not set a x86_64 hwcap to auxv

CharlyYu:
otherwise libc may call some instructions based on the hwcap and make a UNKNOWN_EXCAPTTION


chl337:
That seems like a bug, as _HWCAP_ is processor-dependent. Have you tested it?

CharlyYu:
we can get it from host /proc/$pid/auxv, But it seems we can skip setting it as gvisor did.

Issue #1161, opened 2024-03-19
aarch64: VM panics for " ubuntu bash" 


chl337:
# Issue:

The VM panics while trying to run bash on an Ubuntu image.
The reported reason is: _exception 0x3C on EL1 - Software Breakpoint Instruction exception_.

# Current Build:
* #1133
* #1131
* #1106
* PAN unset in Memcpy

# Test
docker run --rm -it --runtime=quark_d ubuntu bash

---

I suspect a misalignment issue that leads to the wrong opcode, but I am open to suggestions.

shrik3:
have you inspected the bash binary wrt. the faulting PC?

chl337:
With some more debugging:
`
ESR_EL1: 0xF2000001
`
`
PC: 0x40003E599C -> NOTE: it will vary slightly depending on the current build
`
`
PC[opcode]: 0xD400020 => brk #0x1
`
The PC lies in the Kernel, so dumping the kernel points to
`
<_ZN61 ... qkernel..qlib..SysCallID$ ... more mangle ... *fmt..Debug*...>
`
The error is that the current _callId_ is not part of the declared _SysCallID_ definitions, while it is still less than the upper bound _UnkownSyscall_.

I will provide fixies.

shrik3:
breakpoints should be ignored at this point. Apply this if you haven't

Btw. with the patch below the code proceeds to panic at 
```
[INFO] [1/43c0180000|0] (1/1)------get call id sys_io_setup arg0:ffffffffffffff9c
[ERROR] [1/43c0180000|0] NotImplementSyscall syscall SyscallArguments { arg0: ffffffffffffff9c, arg1: e99e0ce0520, arg2: 4, arg3: 200, arg4: 2ac47b5f6b58, arg5: ffffffffffffff9c }
[Print] [1/43c0180000|0] get panic : Some(Sycall not implement syscall is sys_io_setup)
```

```
diff --git a/qkernel/src/interrupt/aarch64/mod.rs b/qkernel/src/interrupt/aarch64/mod.rs
index d1263110..b59fda62 100644
--- a/qkernel/src/interrupt/aarch64/mod.rs
+++ b/qkernel/src/interrupt/aarch64/mod.rs
@@ -212,6 +212,16 @@ pub extern "C" fn exception_handler_el1h_sync(ptregs_addr:usize){
             let far = GetFarEL1();
             MemAbortKernel(ptregs_addr, esr, far, true);
         },
[...]

PR #1162, opened 2024-03-20
Armdev syscall id table consistent

chl337:
Related to #1161
+ 550dbd88c4bef21e444482dc7639db4e16691bee makes the syscall-number validity [check](https://github.com/QuarkContainer/Quark/blob/bf8725fde4ba3e2271a8ca83da0475f3ecdb2658/qkernel/src/lib.rs#L382) consistent.
+ 61336b29328fa14ee2e8b672482fffd1349ee93b produces helpful log information.
+ c122ccb38060d3f03b20332234b03c2e90460cf4 aqv. of linux sys_ni_syscall.

Regarding #1161, it avoids hitting the breakpoint instruction; we should not get there. In the general case, it will log if a non-implemented syscall was called.

### DIFF
--- a/qkernel/src/interrupt/aarch64/mod.rs
+++ b/qkernel/src/interrupt/aarch64/mod.rs
@@ -222,16 +222,16 @@ pub extern "C" fn exception_handler_el1h_sync(ptregs_addr:usize){
                 let ctx_p = ctx_p.cast::<PtRegs>();
                 let ctx = unsafe { &mut *ctx_p };
                 unsafe {
-                     if let Some(opcode) = kernel_def::read_user_opcode(ctx.pc) {
-                         debug!("VM: current-PC: {:#x}, retrieved PC[opcode]:{:#x}.", ctx.pc, opcode);
-                     } else {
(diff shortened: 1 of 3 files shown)


shrik3:
lgtm. It would be nice if you also format the code for the whole function `pub extern "C" fn exception_handler_el1h_sync(ptregs_addr: usize)` There are some misaligned brackets/indentations.

chl337:
> let ctx_p = ptregs_addr as *mut PtRegs;
> let ctx_p = ctx_p.cast::<PtRegs>();
> let ctx = unsafe { &mut *ctx_p };
> ```
> 
> Do you know a better/cleaner/safer way of doing this raw pointer casting?
> 

I am afraid that no.

> It could also be helpful to
> 
>     1. not only printing opcode on the current pc, but also a few context e.g. ranging from pc - 4_4 to pc + 4_4.
> 
>     2. wrapping this into a debugging function, if it deems frequently used.
Indeed, I will see to it one of these days.

shrik3 (approved):

PR #1164, opened 2024-03-22
Separate cuda build from non cuda machine

changhong94:
After this commit, by default make won't build cuda, so there is no cuda dependency by default. In machine with cuda library, need to do make cuda_all, (or cuda_debug, cuda_release etc) to build cuda.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,8 @@ And also install "cargo-xbuild" as below.
 cargo +stable install cargo-xbuild
 ```
 
-And also some extra libraries for compiling GPU module:
+And also some extra libraries for compiling GPU module:  
+(Please note, you can execute following commands even in non nvidia machine to get ability for compiling GPU module.)
 ```sh
 sudo apt-get install libelf-dev nvidia-driver-535
 ```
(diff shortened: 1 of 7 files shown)


shrik3:
is it any possible to build without cuda at all? see #1109 

It's a deal breaker to require the dependencies where the cuda support is not needed.

changhong94:
This is the change to build without cuda by default. It's the solution to
address the issue 1109.

Thanks,
Hong

shrik3:
On Fri Mar 22, 2024 at 5:48 PM CET, changhong94 wrote:
> This is the change to build without cuda by default. It's the solution to
> address the issue 1109.
>
> Thanks,
> Hong
>


I misread your PR in the first place. That's why I deleted that comment from
github. Sorry for the noise / thanks for the patch.

README.md
```diff
@@ -72,7 +72,8 @@ And also install "cargo-xbuild" as below.
 cargo +stable install cargo-xbuild
 ```
 
-And also some extra libraries for compiling GPU module:
+And also some extra libraries for compiling GPU module:
```
QuarkContainer, line 75: If you need to compile with GPU module, please do following.

QuarkContainer (approved):

PR #1167, opened 2024-03-23
add rustfmt configuration

QuarkContainer:


shrik3:
I did the same at the same time haha #1168 , you are so fast! I'll close mine

PR #1177, opened 2024-03-28
Refactoring schedular && task 

yaoxin1995:


qlib/kernel/Kernel.rs
```diff
@@ -822,10 +822,14 @@ impl HostSpace {
...
 
-    pub fn VcpuWait() -> i64 {
-        let mut ret: i64 = 0;
-        HyperCall64(HYPERCALL_VCPU_WAIT, 0, 0, &mut ret as *mut _ as u64, 0);
-        return ret as i64;
+    pub fn VcpuWait() -> TaskId {
+        let mut next: TaskId = TaskId::New(0, 0);
+        HyperCall64(HYPERCALL_VCPU_WAIT, 0, 0, &mut next as *mut _ as u64, 0);
+        
+        assert!(next.PrivateTaskAddr() != 0);
```
shrik3, line 829: Also in a general sense, doesn't this make the system panic, when quark is not configured to run with cc?

Perhaps it's better to wrap this with a cfg item?

qlib/kernel/mod.rs
```diff
@@ -62,12 +62,16 @@ pub static TSC: Tsc = Tsc::New();
...
+
+ 
 pub static LOADER: Singleton<Loader> = Singleton::<Loader>::New();
 pub static WAIT_CONTAINER_FD: AtomicI32 = AtomicI32::new(-1);
+
+// place holder, will never be used by qvisor
 pub static KERNEL_STACK_ALLOCATOR: Singleton<AlignedAllocator> =
     Singleton::<AlignedAllocator>::New();
+// place holder, will never be used by qvisor
+pub static PAGE_MGR: PageMgrRef = PageMgrRef::New();
```
shrik3, line 74: if never used by qvisor, why should these be defined by under qlib/ instead of qkernel?

Are there unresolved reference?

qlib/kernel/quring/uring_async.rs
```diff
@@ -708,16 +745,13 @@ impl AsyncOpsTrait for AsyncAccept {
         /**************************hibernate wakeu **************************/
         // so far the quark hibernate is wakeup by accept.
         // todo: find better to handle this
-        let cc_enabled = SHARESPACE.config.read().EnableCC;
-        if !cc_enabled {
-            if SHARESPACE.reapFileAvaiable.load(Ordering::Relaxed) {
-                ReapSwapIn();
-            }
+        if SHARESPACE.reapFileAvaiable.load(Ordering::Relaxed) {
```
shrik3, line 748: btw what does "reap" mean here? A typo?

qlib/kernel/task.rs
```diff
@@ -165,10 +160,51 @@ impl Drop for Task {
     }
 }
 
+
+// task wrapper on shared memory
+#[repr(C)]
+pub struct TaskWrapper {
```
shrik3, line 166: Perhaps add some documentations on this?  :D

qlib/kernel/task.rs
```diff
@@ -165,10 +160,51 @@ impl Drop for Task {
...
+}
+
+
 #[repr(C)]
 pub struct Task {
     pub context: Context,
+    // address of Task struct
     pub taskId: u64,
+    // address of TaskWrapper struct
+    pub taskWrapperId: u64,
```
shrik3, line 207: Doesn't the new `struct TaskId` already have both fields?

qlib/kernel/task.rs
```diff
@@ -526,14 +566,9 @@ impl Task {
...
-    pub fn TaskId() -> TaskId {
+    pub fn PrivateTaskID() -> u64 {
         let rsp = GetRsp();
-        return TaskId::New(rsp & DEFAULT_STACK_MAST);
-    }
-
-    #[inline(always)]
-    pub fn GetPtr(&self) -> &'static mut Task {
-        return unsafe { &mut *(self.taskId as *mut Task) };
+        return rsp & DEFAULT_STACK_MAST;
```
shrik3, line 571: typo "MAST", but perhaps fix in another PR.

qlib/kernel/taskMgr.rs
```diff
@@ -107,9 +90,10 @@ pub fn IOWait() {
 }
 
 pub fn WaitFn() -> ! {
+    debug!("WaitFn 222");
```
shrik3, line 93: seems like leftover

qlib/task_mgr.rs
```diff
@@ -64,21 +89,19 @@ pub struct Links {
...
+    pub rbx: u64,  // 0x28
+    pub rbp: u64,  // 0x30
+    pub rdi: u64,  // 0x38
+
+    pub fs: u64,   // 0x40
+    pub savefpsate: bool,  // 
     pub X86fpstate: Option<Box<X86fpstate>>,
-    // job queue id
-    pub queueId: AtomicUsize,
+
```
shrik3, line 104: are you sure this change to context struct definition is consistent across the codebase?

qvisor/src/main.rs
```diff
@@ -115,6 +115,7 @@ pub fn AllocatorPrint(_class: usize) -> String {
     return "".to_string();
 }
 
+pub static IS_GUEST: bool = false;
```
shrik3, line 118: perhaps this flag immutable.

qvisor/src/runc/runtime/vm.rs
```diff
@@ -215,18 +211,10 @@ impl VirtualMachine {
         IOURING.SetValue(sharespace.GetIOUringAddr());
 
         unsafe {
-            KERNEL_PAGETABLE.SetRoot(VMS.read().pageTables.GetRoot());
```
shrik3, line 218: What is replacing these lines to set up the kernel pagetable? Perhaps I'm missing something..

shrik3 (commented):
In general, since it's pretty much a breaking change to the task model, I suggest adding some documentations.

qlib/kernel/quring/uring_mgr.rs
```diff
@@ -559,27 +521,23 @@ impl QUring {
...
 
     pub fn UringPush(&self, entry: UringEntry) {
-        {
-            let mut s = SHARESPACE.uringQueue.submitq.lock();
-            s.push_back(entry);
-        }
+        
 
+        let mut entry = entry;
+        loop {
```
chl337, line 533: Wouldn't it be better to break after a certain number of failed attempts (with an "error" state) rather than get stuck in a loop?

qkernel/src/lib.rs
```diff
@@ -561,9 +561,29 @@ pub extern "C" fn rust_main(
         CreateTask(ControllerProcess as u64, ptr::null(), true);
     }
 
+    if id == 2 {
```
chl337, line 565: If vcpu blocks, it should be reflected also [here](https://github.com/QuarkContainer/Quark/blob/df21d540cdc55c0439b2ea0d64b4b5de7de7d79b/qvisor/src/runc/runtime/vm.rs#L277).

qlib/task_mgr.rs
```diff
@@ -254,19 +258,74 @@ impl Scheduler {
...
+impl<T: Clone> Deref for VcpuLocalQueue<T> {
+    type Target = VecDeque<T>;
+
+    fn deref(&self) -> &VecDeque<T> {
+        &self.0
+    }
+}
+
+// impl<T: Clone> DerefMut for VcpuLocalQueue<T> {
```
chl337, line 283: Please remove leftovers.

qkernel/src/lib.rs
```diff
@@ -138,7 +139,7 @@ pub static GLOBAL_ALLOCATOR: HostAllocator = HostAllocator::New();
 
 pub static GUEST_HOST_SHARED_ALLOCATOR: GuestHostSharedAllocator = GuestHostSharedAllocator::New();
 
-
+pub static IS_GUEST: bool = true;
```
chl337, line 142: Should this be a const?

qkernel/src/lib.rs
yaoxin1995: fixed

qlib/kernel/mod.rs
yaoxin1995: After refactoring, Qvisor does not need to use `PAGE_MGR` and  `KERNEL_STACK_ALLOCATOR` anymore.    But the code should still be in qlib because we want to merge cc branch into main.

qlib/kernel/quring/uring_async.rs
yaoxin1995: Not sure, I didn't write the code :D

qlib/kernel/quring/uring_mgr.rs
yaoxin1995: Maybe this is more CPU efficient. But how to handle errors in this case? We may need to be modifying the code all the way up to the guest system call level.  I would say that this should be an optimisation in the future.

qlib/kernel/task.rs
yaoxin1995: Sometimes, process can only access the `Struc Task` (e.g., using get_current_task()).   So `struct Task` also needs to record `taskWrapperId` so that the process can find `taskwrapper`

qlib/task_mgr.rs
yaoxin1995: fixed

qvisor/src/main.rs
yaoxin1995: global static is by default immutable

qvisor/src/runc/runtime/vm.rs
yaoxin1995: After refactoring, the host no longer needs the kernel stack, which is used to allocate tasks on private memory.

qlib/task_mgr.rs
yaoxin1995: This change should only be used in CC, I guess. Other modes should use the original context.  Will add the cfg flag very soon

qlib/kernel/Kernel.rs
yaoxin1995: Switches will be added in a future patch

qlib/kernel/mod.rs
yaoxin1995: Switches will be added later to make cc compatible with upstream

qlib/kernel/task.rs
yaoxin1995: Working on it now

config.json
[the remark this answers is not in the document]
yaoxin1995: fixed

qlib/kernel/task.rs
yaoxin1995: fixed

qlib/kernel/taskMgr.rs
yaoxin1995: fixed

PR #1131, opened 2024-03-03
aarch64: fix mprotect syscall

CharlyYu:
sys_mprotect will reset the page flag, but currently the UserNonAccessable/UserReadOnly/UserReadWrite definition is missing the MT_NORMAL, this will cause a pagefault of the installed page, and we will install another page.

PR #1106, opened 2024-02-19
[WIP] armdev: fix sys_clone

shrik3:
partially resolves #1099 

sys_clone now successfully returns, but the child process hits access flag faults and permission faults. I'm working on it.

PR #1181, opened 2024-04-02
[WIP]Armdev bash vfork

chl337:
Status: non-deterministic behavior
a) Crush - non-recuperable
   P.S. - It must be manually terminated because a loop was added for debugging purposes:
          mappings: `cat /proc/$(pidof quark_d)/maps`

b) Child terminates - parent crushed on SIGSEGV
e.g. `ls`
print_ls_non_deterministic

Test:
1. Get [dash](https://packages.debian.org/buster/dash)
2. extract:
   ` ar p dash-pachage-name.deb data.tar.xz | tar -xJ ./bin/dash`
3. Build:
   I. ./autogen.sh
   II. ./configure
   III. DCCFLAGS="-static -static-libgcc -g" make
4. Dockerfile:
   `FROM ubuntu `
   `WORKDIR /`
   `COPY your_path_to_bash /`
   `ENTRYPOINT ["/dash"]`
5. build image                                   

#### Extra
1. trace dash:
   `strace -f -o log dash -c /usr/bin/ls`
2. Objdump dash:
   ` objdump -dS dash`

### DIFF
--- a/qkernel/src/kernel_def.rs
+++ b/qkernel/src/kernel_def.rs
@@ -403,12 +403,16 @@ pub fn child_clone(userSp: u64) {
     currTask.AccountTaskEnter(SchedState::RunningApp);
     let pt = currTask.GetPtRegs();
 
-    let kernalRsp = pt as *const _ as u64;
+    let kernelRsp = pt as *const _ as u64;
     CPULocal::Myself().SetEnterAppTimestamp(TSC.Rdtsc());
     //currTask.mm.VcpuEnter();
     CPULocal::Myself().SetMode(VcpuMode::User);
     currTask.mm.HandleTlbShootdown();
-    SyscallRet(kernalRsp)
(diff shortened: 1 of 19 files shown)


shrik3:
btw. are there more works to be done with #1106, #1131 and #1133 ?

If we proceed to fix the clone/fork issue here, perhaps let's merge those first. The dev overhead would quite heavy with unmerged dependencies.

cc. @CharlyYu

qkernel/src/kernel_def.rs
```diff
@@ -396,6 +396,11 @@ impl HostSpace {
 
 #[inline]
 pub fn child_clone(userSp: u64) {
+    #[cfg(target_arch = "aarch64")] {
```
shrik3, line 399: I'm not sure if we need to flush TCB here. During the during the context_swap_to the TCB should already be flushed.

PR #1184, opened 2024-04-04
Fix merge bug

123abcpp:
Now cc mode is enabled automatically when cc feature is compiled. The toggle function will be applied after the sev feature is added.

### DIFF
--- a/qkernel/Cargo.toml
+++ b/qkernel/Cargo.toml
@@ -48,9 +48,6 @@ debug = 1
 lto = true
 opt-level = 2
 
-[features]
-cc = []
-
 [profile.release]
 panic = "abort"
 #opt-level = 0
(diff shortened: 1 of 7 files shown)


shrik3:
+    pub fn VcpuWait() -> TaskId {
+        #[cfg (feature = "cc")]
+        if is_cc_enabled(){
+            return Self::VcpuWait_cc();
+        }
+        let mut next: TaskId = TaskId::New(0, 0);
+        HyperCall64(HYPERCALL_VCPU_WAIT, 0, 0, &mut next as *mut _ as u64, 0);
+        assert!(next.PrivateTaskAddr() != 0);
+        assert!(next.SharedTaskAddr() != 0);
+        return next;
+    }

So we could have the "cc" feature enabled, while having ENABLE_CC set to false?
What's the semantic here?

Also, if the ENABLE_CC is meant to be a toggle switch, shouldn't it read value
from config? Right now it can only be toggled via modifying the source code an
recompiling.

yaoxin1995:
The @shrik3 semantics are as follows: quark supports cc functionality when cc features are enabled. When user deploys the pod, he can add the keyword `Confidential-Workload` to the yaml file. This keyword will cause fn `is_cc_enabled` to return true.

PR #1189, opened 2024-04-18
CUDA support: support llama now

justinwang-ubc:


qkernel/Cargo.toml
QuarkContainer, line 1: why we need libc in qkernel? Qkernel is no_std environment and libc doesn't work there

PR #1203, opened 2024-04-30
aarch64: improved log format and clean ups

chl337:
## Includes

* Pretty format for _struct PtRegs_
* Update outdated comment
* Distinguish output log from the kernel
* ~remove dead code~

PR #1207, opened 2024-05-01
aarch64: replace comment as it can lead to confusion

chl337:
Previous comment may lead to the idea that a separate definition is needed for Arm.


chl337:
> I suggest aggregating more such changes before merging.

I will place it together with #1203

PR #1209, opened 2024-05-02
aarch64: refactor on clone creation

chl337:
Refactor for more readability.

### DIFF
--- a/qkernel/src/kernel_def.rs
+++ b/qkernel/src/kernel_def.rs
@@ -401,18 +401,17 @@ pub fn child_clone(userSp: u64) {
     CPULocal::SetKernelStack(currTask.GetKernelSp());
 
     currTask.AccountTaskEnter(SchedState::RunningApp);
-    let pt = currTask.GetPtRegs();
 
-    let kernelRsp = pt as *const _ as u64;
+    let pt_regs_top = currTask.GetPtRegs() as *const _ as u64;
     CPULocal::Myself().SetEnterAppTimestamp(TSC.Rdtsc());
     //currTask.mm.VcpuEnter();
     CPULocal::Myself().SetMode(VcpuMode::User);
     currTask.mm.HandleTlbShootdown();
-    debug!("entering child task: kernelSp/PtRegs @ {:x}", kernelRsp);
+    debug!("entering child task: kernelSp => PtRegs @ {:x}", pt_regs_top);
     #[cfg(target_arch = "x86_64")]
-    SyscallRet(kernelRsp);
+    SyscallRet(pt_regs_top);
     #[cfg(target_arch = "aarch64")]
-    IRet(kernelRsp);
+    IRet(pt_regs_top);
 }
 
 extern "C" {
--- a/qlib/kernel/threadmgr/task_clone.rs
+++ b/qlib/kernel/threadmgr/task_clone.rs
@@ -647,46 +647,53 @@ impl Task {
 
         return Ok(());
     }
-}
 
-pub fn CreateCloneTask(fromTask: &Task, toTask: &mut Task, userSp: u64) {
-    let mut from = fromTask.GetKernelSp();
-    let fromSp = fromTask.GetPtRegs() as *const _ as u64;
+    pub fn copy_PtRegs_to_child(&self, clone_task: &mut Task, user_sp: u64) {
+        let parent_context = self.GetPtRegs() as *const _ as *const u64;
+        let child_context = clone_task.GetPtRegs() as *mut _ as *mut u64;
+        let context_size = core::mem::size_of::<PtRegs>();
 
-    let mut to = toTask.GetKernelSp();
-    let toPtRegs = toTask.GetPtRegs();
-
-    unsafe {
-        while from >= fromSp {
-            *(to as *mut u64) = *(from as *const u64);
-            from -= 8;
-            to -= 8;
+        unsafe {
+            core::ptr::copy_nonoverlapping(parent_context, child_context, context_size);
         }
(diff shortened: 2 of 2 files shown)


qlib/kernel/threadmgr/task_clone.rs
```diff
@@ -647,48 +647,55 @@ impl Task {
...
+
+        self.context.set_para(entry_argument);
+        //
         // 1. set sys_clone return value to 0 to indicate child.
         // 2. set child pc (return addr of current call frame)
         //      to child_clone function
         #[cfg(target_arch = "x86_64")]
         {
-            toPtRegs.rax = 0;
-            toTask.context.place_on_stack(child_clone as u64);
```
shrik3, line 681: Maybe it's a hassle, but I suggest we don't modify x86 code on the arm feature branch.

qkernel/src/kernel_def.rs
```diff
@@ -401,18 +401,17 @@ pub fn child_clone(userSp: u64) {
...
-    let pt = currTask.GetPtRegs();
 
-    let kernelRsp = pt as *const _ as u64;
+    let pt_regs_top = currTask.GetPtRegs() as *const _ as u64;
     CPULocal::Myself().SetEnterAppTimestamp(TSC.Rdtsc());
     //currTask.mm.VcpuEnter();
     CPULocal::Myself().SetMode(VcpuMode::User);
     currTask.mm.HandleTlbShootdown();
-    debug!("entering child task: kernelSp/PtRegs @ {:x}", kernelRsp);
+    debug!("entering child task: kernelSp => PtRegs @ {:x}", pt_regs_top);
```
shrik3, line 410: why the arrow?

qkernel/src/kernel_def.rs
```diff
@@ -401,18 +401,17 @@ pub fn child_clone(userSp: u64) {
...
     #[cfg(target_arch = "x86_64")]
-    SyscallRet(kernelRsp);
+    SyscallRet(pt_regs_top);
```
shrik3, line 412: pt_regs top is a linux thing, and it's a very specific naming (ptrace registers). I don't think it's a good idea to rename the stack pointer to `pt_regs_top` here.

IMHO `pt_regs_top` is even more comfusing: what do you mean by "top" of a data structure? Data grows upwards.

qlib/kernel/threadmgr/task_clone.rs
```diff
@@ -647,48 +647,55 @@ impl Task {
...
+            self.GetPtRegs().rax = pid;
+            self.context.place_on_stack(entry_address);
```
shrik3, line 674: Maybe it's a hassle, but I suggest we don't modify x86 code on the arm feature branch.

qkernel/src/kernel_def.rs
chl337: Stack pointer point to the begging of the PtRegs.

qlib/kernel/threadmgr/task_clone.rs
chl337: That is already a modification of the original.

qkernel/src/kernel_def.rs
chl337: The top of the PtRegs structure:
`let pt_regs_top = currTask.GetPtRegs() as *const _ as u64;`

qlib/kernel/threadmgr/task_clone.rs
```diff
@@ -647,48 +647,55 @@ impl Task {
...
+        to_task.savefpsate = true;
+        to_task.archfpstate = Some(Box::new(
+            from_task.archfpstate.as_ref().unwrap().Fork(),
+        ));
+
+        let kernel_stack_top = to_task.GetPtRegs() as *const _ as u64 - 8;
+        to_task.context.set_sp(kernel_stack_top);
+
+        to_task.context.set_tls(from_task.context.get_tls());
+        to_task.prepare_clone_entry(user_sp);
```
shrik3, line 695: (continue), if you do the indirection and cluster code into something with the naming `prepare_clone_entry`, why not cluster `set_sp`, and `copy_PtRegs_to_child` as well? The later ones also fall in the scope of "prepare_clone_entry"

qkernel/src/kernel_def.rs
shrik3: `GetPtRegs()` gives you the bottom (base), not the "top".

qlib/kernel/threadmgr/task_clone.rs
chl337: In `prepare_clone_entry` was intended to include only what is necessary for the cloned task to enter in the `child_clone`, return address and argument(s). Now, the things I don't like are 1)how I name it and 2)that the rax/x0 are also included instead of having them in `copy_PtRegs_to_child`.

shrik3: but why would the clustering/indirection be necessary in the first place?

chl337: In a previous debugging sessions, notes where made about the how has difficult to read it. That was my motivation. If separation makes it still unreadable, then I would drop this and try to reorganize inside the same call.

shrik3: Well, actually it's not about the motivation, my biggest concern here is that this involves the x86 code. So I'm reluctant to do these stuffs now.

I would suggest we fix the arm side first. We are so behind the main and we need to keep up before any _general_ refactoring (that is, which also changes x86 code)  should be done.

chl337: I see you point. I will drop it.

PR #1215, opened 2024-05-04
aarch64: refactor paging and page fault handler

shrik3:
breaking change: `MemoryManager::MProtect` function now receives another parameter `permOnly`. When passed true, the MProtect routine will NOT touch PTE flags other than permission related ones. This prevents `mprotect` from corrupting PTEs, which causes weird page faults.  Related: #1131

Ideally the `MProtect` function should NEVER touch PTE flags such as `PRESENT` or `TABLE`. This is wrong semantics. However both x86 and aarch64 code relies this to set up forked address spaces, or load new ELFs. We could do some refactoring in the future. 

On PF handling: as suggested by @QuarkContainer , we skip page swapping for now.

### DIFF
--- a/qkernel/src/interrupt/aarch64/fault.rs
+++ b/qkernel/src/interrupt/aarch64/fault.rs
@@ -101,165 +101,102 @@ impl PageFaultErrorCode {
 }
 
 
-pub fn PageFaultHandler(ptRegs: &mut PtRegs, fault_address: u64,
-                        error_code: PageFaultErrorCode) {
+pub fn PageFaultHandler(ptRegs: &mut PtRegs, fault_address: u64, error_code: PageFaultErrorCode) {
+    use self::PageFaultErrorCode as PFEC;
     CPULocal::Myself().SetMode(VcpuMode::Kernel);
     let currTask = Task::Current();
-    // is this call from user
-    let fromUser: bool = error_code
-        .contains(PageFaultErrorCode::USER_MODE);
-    //
-    // NOTE: ATM only 0x0000xx...x address-space is used.
-    //
+    let fromUser: bool = error_code.contains(PFEC::USER_MODE);
     let ttbr = CurrentUserTable();
 
     if !SHARESPACE.config.read().CopyDataWithPf && !fromUser {
-        error!("VM: PageFault in kernel FAR: {:#x}, TTBR: {:#x}, PtRegs: {:#x}, error_code: {:#x}",
-               fault_address, ttbr, ptRegs, error_code);
-       //backtracer::trace(ptRegs.pc, ptRegs.get_stack_pointer(), ptRegs.rbp, &mut |frame| {
-       //      print!("pagefault frame is {:#x?}", frame);
-        //     true
-        // });
+        error!(
+            "VM: PageFault in kernel FAR: {:#x}, TTBR: {:#x}, PtRegs: {:#x}, error_code: {:#x}",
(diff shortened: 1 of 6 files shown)


chl337:
Do the new changes affect (solve) the current issue regarding the allocator? (I can have a better look and test it first on Monday, but if it can not wait go on and merge it)

shrik3:
On Sat May 4, 2024 at 8:01 AM CEST, chl337 wrote:
> Do the new changes affect (solve) the current issue regarding the
> allocator? (I can have a better look and test it first on Monday, but
> if it can not wait go on and merge it)


unfortunately no.

But it should solve at least one potential issue, I put my reasoning
here and also propose future changes. cc. @QuarkContainer

before the change, the MProtect (the low level ones from MM and
Pagetable mods, which the sys_mprotect uses) is implemented in a way
that, it takes some flags returned from PageOpts::* and overwrite
everything except for the address field of the PTE. I don't really like
this approach, mprotect should do mprotect, nothing else. What's the
point of having mprotect writting to pteflags.present or
pteflags.accessed?

for aarch64 specific:

https://sys.shrik3.com/aarch64/memory/table_descriptor.html

only flags like UXN, PXN, an AP fields are supposed to be modified by
MProtect.

So the previous MProtect simply takes a predefined flag wrt. the
requested permission change. And writes to PTE except for [47:12]  (the
physical addr bits) regard less of the previous state.

Take this for example:

```rust

pub fn UserReadOnly() -> Self {
    return PageOpts(
            PageTableFlags::USER_ACCESSIBLE |
            PageTableFlags::READ_ONLY |
            );
}

```

Then the "user read only" page will lose all attributes other than these
two. Which caused pagefaults obviously. So the workaround __was__ to do


```rust

pub fn UserReadOnly() -> Self {
    return PageOpts(
            PageTableFlags::USER_ACCESSIBLE |
            PageTableFlags::READ_ONLY |
            PageTableFlags::VALID |
            PageTableFlags::MT_NORMAL |
            PageTableFlags::ACCESSED
            );
}

```

the same goes for x86 code:

```rust

pub fn UserReadOnly() -> Self {
    return PageOpts(PageTableFlags::PRESENT | PageTableFlags::USER_ACCESSIBLE);
}
```

But this is not really solving the issue.

---

More on this, the reason I'm giving MProtect() an additional flag is to
make code compatible: as said, Quark "misuse" this overwritting
"feature" of MProtect to do things like forking address spaces or
setting up new address spaces for newly loaded programs. This should be
changed but I don't want do do it here yet.



    cheers.

chl337:
The issue with MProtect is not new. The previous PfH only masked one of the problems caused by it (I am conscious that was not clean). The real question is, should a fix be provided now, or after rebaseing with main🤷🏻‍♂️

shrik3:
> The issue with MProtect is not new. The previous PfH only masked one of the problems caused by it (I am conscious that was not clean).

It wasn't clean, but other issues caused by mprotect do not necessarily land you in the PF path. So restoring VALID and TABLE flags in the PF was the only thing you could do there.

by all means, the PF handler should not clean up the mess caused by mprotect or whatever. Let PFs be PFs!

> should a fix be provided now, or after rebaseing with main🤷🏻‍♂️

This changes do not affect main at all, and the crosscutting code is minimal. I think we need a temporal fix (like this) for aarch64 now, because it seems overwriting pte flags cause more problems in aarch64 than x86

chl337 (approved):

PR #1217, opened 2024-05-07
aarch64: implement signal delivery

shrik3:
**[updated!]** There are some todos, see commit messages. Anyways for the scope of this PR I'll stop here.

we have the basic stuffs running without issues now. try container `ubuntu bash`
image

### DIFF
--- a/qlib/kernel/SignalDef.rs
+++ b/qlib/kernel/SignalDef.rs
@@ -346,6 +346,7 @@ pub const UC_SIGCONTEXT_SS: u64 = 2;
 pub const UC_STRICT_RESTORE_SS: u64 = 4;
 
 // https://elixir.bootlin.com/linux/latest/source/include/uapi/asm-generic/ucontext.h#L5
+#[cfg(target_arch = "x86_64")]
 #[repr(C)]
 #[derive(Debug, Copy, Clone, Default)]
 pub struct UContext {
@@ -356,6 +357,7 @@ pub struct UContext {
     pub Sigset: u64,
 }
 
+#[cfg(target_arch = "x86_64")]
 impl UContext {
     pub fn New(ptRegs: &PtRegs, oldMask: u64, cr2: u64, fpstate: u64, alt: &SignalStack) -> Self {
         return Self {
@@ -367,6 +369,47 @@ impl UContext {
         };
     }
 }
+
+#[cfg(target_arch = "aarch64")]
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct UContext {
+    pub Flags: u64,
+    pub Link: u64,
+    pub Stack: SignalStack,
+    pub Sigset: u64,
+    __pad: [u8; (1024 / 8) - core::mem::size_of::<u64>()],
+    pub MContext: SigContext,
+}
+
+#[cfg(target_arch = "aarch64")]
+impl Default for UContext {
+    fn default() -> Self {
+        Self {
+            Flags: 0,
+            Link: 0,
+            Stack: SignalStack::default(),
+            Sigset: 0,
+            __pad: [0; (1024 / 8) - core::mem::size_of::<u64>()],
+            MContext: SigContext::default(),
+        }
+    }
+}
+
+#[cfg(target_arch = "aarch64")]
+impl UContext {
+    pub fn New(ptRegs: &PtRegs, oldMask: u64, cr2: u64, fpstate: u64, alt: &SignalStack) -> Self {
+        return Self {
+            Flags: 0,
+            Link: 0,
+            Stack: alt.clone(),
+            Sigset: 0,
+            __pad: [0u8; (1024 / 8) - core::mem::size_of::<u64>()],
+            MContext: SigContext::New(ptRegs, oldMask, cr2, fpstate),
+        };
+    }
+}
+
 
 // https://elixir.bootlin.com/linux/latest/source/arch/x86/include/uapi/asm/sigcontext.h#L284
 #[cfg(target_arch = "x86_64")]
--- a/qlib/kernel/SignalDef.rs
+++ b/qlib/kernel/SignalDef.rs
@@ -398,18 +398,21 @@ impl Default for UContext {
 
 #[cfg(target_arch = "aarch64")]
 impl UContext {
-    pub fn New(ptRegs: &PtRegs, oldMask: u64, cr2: u64, fpstate: u64, alt: &SignalStack) -> Self {
+    pub fn New(ptRegs: &PtRegs, oldMask: u64, fault_address: u64, fpstate: u64, alt: &SignalStack) -> Self {
         return Self {
             Flags: 0,
             Link: 0,
             Stack: alt.clone(),
             Sigset: 0,
             __pad: [0u8; (1024 / 8) - core::mem::size_of::<u64>()],
-            MContext: SigContext::New(ptRegs, oldMask, cr2, fpstate),
+            MContext: SigContext::New(ptRegs, oldMask, fault_address, fpstate),
         };
     }
 }
 
+// HACK. TODO Refactor this as soon as we have proper VDSO perser.
+#[cfg(target_arch = "aarch64")]
+pub const VDSO_OFFSET_SIGRETURN: u64 = 0x1120;
 
 // https://elixir.bootlin.com/linux/latest/source/arch/x86/include/uapi/asm/sigcontext.h#L284
 #[cfg(target_arch = "x86_64")]
@@ -460,6 +463,7 @@ pub struct SigContext {
 	pub sp: u64,
 	pub pc: u64,
 	pub pstate: u64,
+    pub oldmask: u64, /* should we do it here? */
 	/* 4K reserved for FP/SIMD state and future expansion */
 	pub __reserved: [u8; 4096],
 }
@@ -504,8 +508,13 @@ impl SigContext {
 impl Default for SigContext {
     fn default() -> Self {
         Self {
+            fault_address: 0,
+            regs: [0u64; 31],
+            sp: 0,
+            pc: 0,
+            pstate: 0,
+            oldmask: 0,
             __reserved: [0u8;4096],
-            ..Default::default()
         }
     }
 }
@@ -523,16 +532,17 @@ impl fmt::Debug for SigContext {
     }
 }
 
-//TODO  how to set sigcontext
+//TODO support fpstate
 #[cfg(target_arch = "aarch64")]
 impl SigContext {
-    pub fn New(ptRegs: &PtRegs, oldMask: u64, cr2: u64, fpstate: u64) -> Self {
+    pub fn New(ptRegs: &PtRegs, oldMask: u64, fault_address: u64, _fpstate: u64) -> Self {
         return Self {
-            fault_address: 0,
+            fault_address: fault_address,
             regs: ptRegs.regs.clone(),
             pc: ptRegs.pc,
             sp: ptRegs.sp,
             pstate: ptRegs.pstate,
+            oldmask: oldMask,
             ..Default::default()
         };
     }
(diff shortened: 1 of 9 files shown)


shrik3:
if i may add, for the sake of simplicity, the fpstate is completely ignored in
the signal context. This is left fot the readers as practice. (这都是作者的良苦
用心)

also with the current task/context structs (and their compositions) it's hacky
to plug in architectural extensions such as aarch64 SIMD (SME and SVE). they are
ignored as well.

chl337:
This is the result of the [test](https://paste.sr.ht/~shrik3/980abc7b3081729cb6c4c685a6956b36a30ba487) directly on the machine:
scrshot_2024-05-07T09:52:46

Is this what to be expected?

chl337:
The _struct ucontext_ is different between x86_64 and [arm64](https://elixir.bootlin.com/linux/latest/source/arch/arm64/include/uapi/asm/ucontext.h#L22) in Linux? Could this cause cause confusion to glibc?

[This](https://paste.sr.ht/~ch3lt/0be2e678b04d7192e1beb877de59eae5f8838782) implementation breaks _SignalReturn()_.

shrik3:
> This is the result of the [test](https://paste.sr.ht/~shrik3/980abc7b3081729cb6c4c685a6956b36a30ba487) directly on the machine: scrshot_2024-05-07T09:52:46
> 
> Is this what to be expected?
> 
> On Quark: scrshot_2024-05-07T09:58:45

> Is this what to be expected?

Yes. in the test I have no sanity check whether the vdso `__kernel_rt_sigreturn` is a null pointer. If you pass 0 to sigaction and set restorer to true then you will get funky things.

shrik3:
> The _struct ucontext_ is different between x86_64 and [arm64](https://elixir.bootlin.com/linux/latest/source/arch/arm64/include/uapi/asm/ucontext.h#L22) in Linux? Could this cause cause confusion to glibc?
> 
> [This](https://paste.sr.ht/~ch3lt/0be2e678b04d7192e1beb877de59eae5f8838782) experimental (copy-paste from linux) implementation breaks _SignalReturn()_.

good point. let me see

chl337:
> known issue:
> 
>     1. could panic if the user does not set a working trampoline for the sigreturn. Interestingly the libc `signal()` does not use vdso `__kernel_rt_sigreturn` actively. Perhaps we are missing some stuffs.
> 
How can this behavior be reproduced?

shrik3:
> > known issue:
> > ```
> > 1. could panic if the user does not set a working trampoline for the sigreturn. Interestingly the libc `signal()` does not use vdso `__kernel_rt_sigreturn` actively. Perhaps we are missing some stuffs.
> > ```
> 
> How can this behavior be reproduced?

see my example, replace the sigaction stuffs with one `signal(signo, handler)`. The return trampoline will not be correctly called.

chl337:
> > > known issue:
> > > ```
> > > 1. could panic if the user does not set a working trampoline for the sigreturn. Interestingly the libc `signal()` does not use vdso `__kernel_rt_sigreturn` actively. Perhaps we are missing some stuffs.
> > > ```
> > 
> > 
> > How can this behavior be reproduced?
> 
> see my example, replace the sigaction stuffs with one `signal(signo, handler)`. The return trampoline will not be correctly called.

Thanks and sorry for the noise, I am a totally noob on the subject.

qlib/kernel/threadmgr/task_signals.rs
```diff
@@ -1263,9 +1263,115 @@ impl Task {
...
+        regs.pc = sigAct.handler;
+        // TODO set proper value for PSTATE
+        regs.pstate = 0x0;
+        // parameters to the handler
+        // handler(int signo, siginfo_t *info, void *context)
+        debug!(">>> deliver signal: restorer is {:#X}", sigAct.restorer);
+        regs.regs[0] = signo;
+        regs.regs[1] = sigInfoAddr;
+        regs.regs[2] = sigCtxAddr;
+        regs.regs[30] = sigAct.restorer;
```
chl337, line 1332: If I understand _[sygaction(2)](https://man7.org/linux/man-pages/man2/sigaction.2.html)_, for x86_64 the _.restore_ field is valid if the _SA_RESTORE_ flag is set. For aarch64, that uses the vDSO, should it be taken as default?

shrik3: it's not the kernel's responsibility to set the restorer addr to vdso function  here. The user library parses vdso, and give the address to sigaction.

shrik3:
> > The _struct ucontext_ is different between x86_64 and [arm64](https://elixir.bootlin.com/linux/latest/source/arch/arm64/include/uapi/asm/ucontext.h#L22) in Linux? Could this cause cause confusion to glibc?
> > [This](https://paste.sr.ht/~ch3lt/0be2e678b04d7192e1beb877de59eae5f8838782) experimental (copy-paste from linux) implementation breaks _SignalReturn()_.
> 
> can confirm, I think the problem is with stack unwinding.

fixed. The problem is with UContext Default constructor. (it has infinite recursion)

I don't know exactly why but patterns like this should be avoided.

```
+impl Default for UContext {
+    fn default() -> Self {
+        Self {
+            __pad: [0; (1024/8) - core::mem::size_of::<u64>()],
+            ..Default::default()                                                          <= this causes recursion into Self::default
+        }
+    }
+}
```

chl337:
> > > The _struct ucontext_ is different between x86_64 and [arm64](https://elixir.bootlin.com/linux/latest/source/arch/arm64/include/uapi/asm/ucontext.h#L22) in Linux? Could this cause cause confusion to glibc?
> > > [This](https://paste.sr.ht/~ch3lt/0be2e678b04d7192e1beb877de59eae5f8838782) experimental (copy-paste from linux) implementation breaks _SignalReturn()_.
> > 
> > 
> > can confirm, I think the problem is with stack unwinding.
> 
> fixed. The problem is with UContext Default constructor. (it has infinite recursion)
> 
> I don't know exactly why but patterns like this should be avoided.
> 
> ```
> +impl Default for UContext {
> +    fn default() -> Self {
> +        Self {
> +            __pad: [0; (1024/8) - core::mem::size_of::<u64>()],
> +            ..Default::default()                   <= this causes recursion into Self::default
> +        }
> +    }
> +}
> ```

Oh, yes I see.
https://github.com/rust-lang/rust-clippy/issues/8609

qlib/kernel/threadmgr/task_signals.rs
chl337: Then I don't know. This was the most obvious arch. difference in the manual.

shrik3: > it's not the kernel's responsibility to set the restorer addr to vdso function here. The user library parses vdso, and give the address to sigaction.

I stand corrected. the kernel should point the trampoline to vdso symbols if restorer is not present. Will fix.

chl337: I believe _xmas-elf_ crate can be used to read the symbol table of the vDSO, and get the symbol. Ping me if I should provide a patch.

shrik3: I have a temporary workaround only for the signals. 

I don't want to add extensive VDSO support in this series. Could you perhaps implement a module that returns address from loaded user vdso by symbol name (str) ?  Please have another PR if you like.

Something like linux kernel `arch/arm64/include/asm/vdso.h` would be nice. (Perhaps store the VDSO base in task.mm)

```
#include <generated/vdso-offsets.h>
#define VDSO_SYMBOL(base, name)						   \
({									   \
	(void *)(vdso_offset_##name - VDSO_LBASE + (unsigned long)(base)); \
})
```

Also take a look at 

```
pub fn GetVDSOParamPageAddr() -> u64 {
    return TIME_KEEPER.read().params.GetParamPageAddr();
}
```

from `qlib/kernel/kernel/timer/mod.rs`. Since VDSO is not only about time, I'd like to have relevant code/data in in a different place.

shrik3: I have an additional field in task.mm called `UserVDSOBase`. This value is loaded when a user task is loaded.

chl337: I 'll look into it.

qlib/kernel/SignalDef.rs
```diff
@@ -480,16 +532,17 @@ impl fmt::Debug for SigContext {
...
 
-//TODO  how to set sigcontext
+//TODO support fpstate
 #[cfg(target_arch = "aarch64")]
 impl SigContext {
-    pub fn New(ptRegs: &PtRegs, oldMask: u64, cr2: u64, fpstate: u64) -> Self {
+    pub fn New(ptRegs: &PtRegs, oldMask: u64, cr2: u64, _fpstate: u64) -> Self {
         return Self {
-            fault_address: 0,
+            fault_address: cr2,
```
chl337, line 540: Could you please adjust the naming - [suggested patch](https://paste.sr.ht/blob/1d58b06e9669ed4877dba13fae8a84786713861a)

shrik3: that was a leftover, I'll fix, thx

CharlyYu:
I still get an error when run ubuntu bash


and this is the log:
image

The code is in the branch  of armdev_signals+pr+wip and  I get qkernel_d.bin and quark_d by `make debug`.

CharlyYu:
And It seems the signal is handled repeatedly.

image

chl337:
> And It seems the signal is handled repeatedly.
> 
> image

Can you show the last instructions before the first SIGSEGV (signal 11)?

CharlyYu:
> Can you show the last instructions before the first SIGSEGV (signal 11)?

CharlyYu:
It seems a page fault with an inappropriate address.

chl337:
> git log: 

Can you also apply #1220 and test again? Also do a _make clean_ first.

CharlyYu:
Same error, is it because I didn't cherry-pick codes in #1212 or other prs? can you push a branch of your codes that can run the ubuntu bash?

chl337:
> Same error, is it because I didn't cherry-pick codes in #1212 or other prs? 
My tree: #1212 + #1215 + #1217 + #1220

can you push a branch of your codes that can run the ubuntu bash?

I can do it first in ~1.5 hours, sorry.

shrik3:
@CharlyYu  sorry for late reply.

https://github.com/shrik3/Quark/tree/armdev_dev_base -> for you only 

image

the take a look at the commit log and throw away my local changes (prefixed with local) that you don't want.

#1220 is not necessary for ubuntu bash to run. Also note that the quark is broken on the newest ubuntu container image (both on main and us). Try using something older. Also you could try `busybox sh`

chl337:
> It seems a page fault with an inappropriate address.

@CharlyYu That seems to me as a context switch issue... If not already solved, try to print the state during switches (use #1203 for pretty print)

chl337:
On trying Redis I get the following error:
`bad g in signal handler`
which for me is pretty cryptic. The only reference I found online is on [Go runtime](https://go.googlesource.com/go/+/go1.15.3/src/runtime/signal_unix.go#919). AFAIK Redis is supported on main branch (cannot tested myself though). 

Reproduce:
`docker run --rm -it -p 6379:6379 --runtime=quark_dlocal --name redis_test redis`

shrik3:
> On trying Redis I get the following error: `bad g in signal handler` which for me is pretty cryptic. The only reference I found online is on [Go runtime](https://go.googlesource.com/go/+/go1.15.3/src/runtime/signal_unix.go#919). AFAIK Redis is supported on main branch (cannot tested myself though).
> 
> Reproduce: `docker run --rm -it -p 6379:6379 --runtime=quark_dlocal --name redis_test redis`

this: https://go.googlesource.com/go/+/go1.15.3/src/runtime/signal_unix.go#378

please grep sys_rt_sigaction in the log. What's sigaction (the restorer specifically)?

chl337:
I have attached the log here.

chl337:
Alright, I will open a separate issue for this.

chl337:
On Fri May 10, 2024 at 11:15 AM CEST, Tianhao Wang wrote:
> > On trying Redis I get the following error: `bad g in signal handler` which for me is pretty cryptic. The only reference I found online is on [Go runtime](https://go.googlesource.com/go/+/go1.15.3/src/runtime/signal_unix.go#919). AFAIK Redis is supported on main branch (cannot tested myself though).
> > 
> > Reproduce: `docker run --rm -it -p 6379:6379 --runtime=quark_dlocal --name redis_test redis`
>
> this: https://go.googlesource.com/go/+/go1.15.3/src/runtime/signal_unix.go#378
>
> please grep sys_rt_sigaction in the log. What's sigaction (the restorer specifically)?

Alright, I will open a new issue for this.

PR #1220, opened 2024-05-08
[wip] force SIGILL on illegal instr from EL0

shrik3:
```
4 SIGILL Illegal instruction. 
The program contained some machine code the CPU can't understand.
```

some userspace program insert illegal exception and listen on (SIGILL 4). This could be some sort of software breakpoint.
e.g. nginx uses this (I don't know why .... )

Anyways, EL0 exceptions should not cause the kernel to panic. Note that this commit is not complete. (some code should be added after calling MainRun ... )


shrik3:
drop for rebasing

PR #1221, opened 2024-05-08
Cc test1

yaoxin1995:


config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "DebugLevel"    : "Info",
+  "DebugLevel"    : "Trace",
```
chl337, line 2: Place drop this change if not necessary.

qlib/linux_def.rs
```diff
@@ -3037,6 +3038,238 @@ impl MemoryDef {
     pub const KERNEL_END_P2_ENTRY: usize = (Self::PHY_UPPER_ADDR / Self::ONE_GB) as usize; //512
 }
 
+
+#[cfg(feature = "cc")]
+pub struct MemoryDef {}
```
chl337, line 2933: Why declared twice if it does not differ from the no-cc?

qlib/linux_def.rs
```diff
@@ -3039,6 +3044,193 @@ impl MemoryDef {
...
+    }
+
+
+    pub const fn rdma_global_shared_hva () -> u64 {
+        return Self::RDMA_GLOBAL_SHARE_OFFSET;
+    }
+
+    //======================== shared ====================//
+
+    pub const fn phy_upper_gpa () -> u64 {
```
shrik3, line 3175: if these stay constant and there is no runtime calculation, why do we need getter functions?

qlib/linux_def.rs
```diff
@@ -3037,6 +3038,238 @@ impl MemoryDef {
     pub const KERNEL_END_P2_ENTRY: usize = (Self::PHY_UPPER_ADDR / Self::ONE_GB) as usize; //512
 }
 
+
+#[cfg(feature = "cc")]
+pub struct MemoryDef {}
+
+#[cfg(feature = "cc")]
+impl MemoryDef {
```
chl337, line 2935: This seems as a lot of reduplication from the non-cc case. Did you try to cluster only the parts that differ? That would help readability.

qlib/pagetable.rs
```diff
@@ -604,15 +605,17 @@ impl PageTables {
  
         let mut curAddr = start;
         let pt: *mut PageTable = MemoryDef::gpa_to_hva(self.GetRoot()) as *mut PageTable;
+
         unsafe {
             let mut p4Idx = VirtAddr::new(curAddr.0).p4_index();
             let mut p3Idx = VirtAddr::new(curAddr.0).p3_index();
-
+
```
chl337, line 612: Empty space! :)

shrik3:
Just for sanity checking: does the program work the same _WITHOUT_ cc enabled?

qkernel/src/kernel_def.rs
```diff
@@ -235,7 +237,7 @@ pub fn HyperCall64(type_: u16, para1: u64, para2: u64, para3: u64, para4: u64) {
 #[inline(always)]
 pub fn HyperCall64(type_: u16, para1: u64, para2: u64, para3: u64, para4: u64) {
     let vcpuid = GetVcpuId();
-    let share_para_page  = MemoryDef::HYPERCALL_PARA_PAGE_OFFSET as *mut ShareParaPage;
+    let share_para_page  = MemoryDef::hcall_page_gpa() as *mut ShareParaPage;
```
shrik3, line 240: I forgot to ask in the previous PR. But are you sure the HyperCall interface is synchronized? Otherwise this hypercall param shared buffer should be on a per-cpu (or thread) basis.

yaoxin1995:
> Just for sanity checking: does the program work the same _WITHOUT_ cc enabled?
@shrik3  Yes, it works without cc enabled.

qkernel/src/lib.rs
```diff
@@ -159,12 +161,23 @@ pub fn SingletonInit() {
         vcpu::VCPU_COUNT.Init(AtomicUsize::new(0));
         vcpu::CPU_LOCAL.Init(&SHARESPACE.scheduler.VcpuArr);
         InitGs(0);
+
         KERNEL_PAGETABLE.Init(PageTables::Init(CurrentKernelTable()));
-        //init fp state with current fp state as it is brand new vcpu
-        FP_STATE.Reset();
+        // unsafe {
```
shrik3, line 166: leftovers :cat2:

qlib/linux_def.rs
```diff
@@ -2983,47 +2984,51 @@ impl MemoryDef {
...
+    // heap
+    pub const HOST_INIT_HEAP_OFFSET: u64 = MemoryDef::NVIDIA_START_ADDR + Self::NVIDIA_ADDR_SIZE;
+    pub const HOST_INIT_HEAP_SIZE: u64 = 5 * Self::ONE_GB;
+    pub const HOST_INIT_HEAP_END: u64 = Self::HOST_INIT_HEAP_OFFSET + Self::HOST_INIT_HEAP_SIZE;
+
     // memory layout
     // PHY_LOWER_ADDR: qkernel image 512MB
     pub const QKERNEL_IMAGE_SIZE: u64 = 512 * Self::ONE_MB;
+
+    //==================================================== private ==============================================//
```
shrik3, line 3002: Please don't use this LOOOOOONG separators. Limit the length to 72 or so, otherwise it will wrap very ugly.

qlib/linux_def.rs
chl337: Oh, I was looking at older commits.

qlib/linux_def.rs
shrik3: IMHO having getters is not solving the real issue, if not making it worse. Because all the defs constant are still in the same place i.e. MemDefs. If separating host and guest memory defs is the case then we should consider refactoring MemoryDefs first.

qlib/linux_def.rs
yaoxin1995: resolved

qkernel/src/lib.rs
yaoxin1995: removed

PR #1223, opened 2024-05-08
[THIS IS A TEST] WILL DELETE SOON

shrik3:
I'm trying to make sense of the diff between current armdev and main.


chl337:
It doesn't look that bad😅

shrik3:
> It doesn't look that bad😅

better than I expected! But I'd like to take chance to redeem some bad commits :D

qlib/kernel/memmgr/pma.rs
```diff
@@ -52,7 +52,9 @@ impl PageMgr {
     }
 
     #[cfg(target_arch = "aarch64")]
-    pub fn Clear(&self) {}
+    pub fn Clear(&self) {
+        // todo!("PageMgr::Clear not implemented for aarch64");
```
chl337, line 56: Do you want to open a issue for these todos? Otherwise they may pass unnoticed.

shrik3: > Do you want to open a issue for these todos? Otherwise they may pass unnoticed. Or should I make a list, including other dummy placeholders?

that would be nice. A grep would work I guess. (if we are writing good comments). Let's do this after the rebase, hopefully we finish today.

PR #1231, opened 2024-05-11
aarch64 support

shrik3:
# [UPDATED] one last round of RFC! 

**Please manifest your suggestions. Otherwise I'll rebase the commit log and ping for a x86 regression test, then hit the merge button.**

**Instructions** on testing & suggesting patches:

- take this PR branch (from my personal repo), or:
    `git fetch origin pull/1231/head:aarch64_test` then checkout to `aarch64_test`
- test on both x86 and aarch64, see what's broken. 
- for further patches, you can simply open PRs or send me patches, I'll manually apply them.

**current state**
quark is working on both x86 and aarch64. We have some regressions in aarch64 [^reg]. Whether there is regression on x86 is unknown, please help with testing!

there are a few items on the checklist, otherwise we are ready to merge, cheers.

**checklist**

- [x] fix runtime errors on aarch64 (introduced by new commits from main)
- [x] fix compilation errors on x86
- [x] pending patches (I hand-picked the commits into this PR, the original PRs are closed)
    - [x] #1234 
    - [x] #1237
- [x] fix vdso build (for both archs)
    - [x] it seems some sources are missing
    - [x] remove vdso objects from the git tracking.
- [x] fix makefile (regarding pinning toolchains) for both archs
- [x] test x86 for regression
- [x] update readme for aarch64 support
- [x] cleanups (formats, linter, comments, documentations, etc.)
     - Note: I try to fix as many warning as possible, some of them however require more refactoring to fix due to conditional compiling. 
- [x] if you like, please review this PR. (well, it's a lot, I know!). One key point is to check if there are leftover aarch64 specific things that are not wrapped in a `#[cfg(...)]` block.
- [x] review merge commit message (see below) and collect **Signoffs** (please feel free to append)

**Draft merge commit message**

```
This pull provides preliminary aarch64 support.

CharlyYu (17):
      aarch64: set vcpu init by kvm_bindings
      aarch64: fix elf loader problems
      aarch64: add kernel pagetable identical mapping
      aarch64: add vdso of arm64
      aarch64: support tls for aarch64
      aarch64: rename GetRsp to GetCurrentKernelSp
      aarch64: fix address issue
      aarch64: specify copy dst len with src len
      aarch64: distinguish arch for LibcStat
      aarch64: add SyscallTable for aarch64
      aarch64: distinguish arch when set auxv
      aarch64: fix substraction overflow in NewMmapLayout
      aarch64: support EnterUser for aarch64
      aarch64: add CopyPageUnsafe asm codes for arm64
[...]

PR #1234, opened 2024-05-14
fix aarch64 compile warning

QuarkContainer:


shrik3:
thanks, this will be rebased into #1231 to merge at once.

shrik3:
Applied in #1231

PR #1237, opened 2024-05-15
aarch64: fix arch specific epoll event issue

CharlyYu:
The epoll_event defined in linux is different on aarch64 than on x86_64, on x86_64, it is __attribute__((packed)) so the size of it is 12, but on aarch64 there is no such attribute, makes its size 16, and we have to add padding between Events and the data.

With this PR, nginx could run on arm64

image


shrik3:
Thanks, I'll rebase this patch onto #1231

shrik3:
Applied in #1231

Issue #1244, opened 2024-05-17
aarch64: build broken 


chl337:
main does not build on aarch:
```
error[E0061]: this method takes 5 arguments but 4 arguments were supplied
    --> src/interrupt/aarch64/fault.rs:177:35
     |
177  |             let res = currTask.mm.InstallPageLocked(currTask, &vma, pageAddr, &range);
     |                                   ^^^^^^^^^^^^^^^^^---------------------------------- an ar
gument of type `bool` is missing
     |
note: method defined here
    --> src/qlib/kernel/memmgr/mm.rs:1049:12
```

The issues it seems trivial, but I am not sure how it was introduced.

HEAD: 1d3fbc23

chl337:
Also, after building with #1245, I am noticing some regression on aarch64.

### Test
``` docker run --rm -it runtime=quark_d busybox sh``` + ``` ls ```
@shrik3 @CharlyYu can you confirm?

shrik3:
On Fri May 17, 2024 at 12:37 PM CEST, chl337 wrote:
> Also, after building with #1245, I am noticing some regression on aarch64.
>
> ### Test
> ``` docker run --rm -it runtime=quark_d busybox sh``` + ``` ls ```
> @shrik3 @CharlyYu can you confirm?

as expected.

please follow up @1240 to apply the same optimization for aarch64 in #1245.
Adding the needs_write flag is not the whole story.

chl337:
On Fri May 17, 2024 at 12:46 PM CEST, Tianhao Wang wrote:
> On Fri May 17, 2024 at 12:37 PM CEST, chl337 wrote:
> > Also, after building with #1245, I am noticing some regression on aarch64.
> >
> > ### Test
> > ``` docker run --rm -it runtime=quark_d busybox sh``` + ``` ls ```
> > @shrik3 @CharlyYu can you confirm?
>
> as expected.
>

> please follow up @1240 to apply the same optimization for aarch64 in #1245.
I am not sure the above is the right link:)

shrik3:
> Also, after building with #1245, I am noticing some regression on aarch64.
> ### Test
> 
> ` docker run --rm -it runtime=quark_d busybox sh` + `ls` @shrik3 @CharlyYu can you confirm?

I may have a fix on that but what's your error here exactly?

chl337:
> > Also, after building with #1245, I am noticing some regression on aarch64.
> > ### Test
[...]

PR #1245, opened 2024-05-17
aarch64: needWrite-flag missing when installing page on PF

chl337:
Fixes issue #1244.

**Please** test first if you can reproduce #1244, and then apply the commit.


shrik3:
On Fri May 17, 2024 at 12:32 PM CEST, chl337 wrote:
> Fixes issue #1244.

No, as I pointed out in that issue, this is not a regression fix. This is a
followup we need to do on the aarch64 side. Please go through #1240 and apply
the optimizations accordingly.

> **Please** test first if you can reproduce #1244, and then apply the commit.

This is expected. In short InstallPageLocked is not installing the page
correctly on aarch64 because we are missing stuffs there.

> -- Commit Summary --
>
>   * aarch64: needWrite-flag missing when installing page on PF

Please apply my suggestions above and rename the commit(s).

Ping me if you need help :D

shrik3 (approved):

PR #1260, opened 2024-05-19
workaround for ARM readlinkat bug

QuarkContainer:


shrik3:
Thanks, I'll test soon

shrik3 (approved):

PR #1268, opened 2024-05-21
Backtrack

QuarkContainer:
I don't know how to install llvm-addr2line so I use addr2line instead.

### DIFF
--- a/qkernel/src/lib.rs
+++ b/qkernel/src/lib.rs
@@ -713,6 +713,16 @@ fn panic(info: &PanicInfo) -> ! {
         print!("panic occurred but can't get location information...");
     }
 
+    qlib::backtracer::trace(
+        GetCurrentKernelIp(),
+        GetCurrentKernelSp(),
+        GetCurrentKernelBp(),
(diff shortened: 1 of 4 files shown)


shrik3:
> I don't know how to install llvm-addr2line so I use addr2line instead.

with ubuntu I think you can simply install the llvm package. The llvm-addr2line tool would be included.

I tested both, the llvm-addr2line is much faster (almost immediate feedback) than the old addr2line tool.

shrik3 (approved):
please use the "squash and merge" option, as this PR contains already-merged commits.

QuarkContainer:
@shrik3 for the "install the llvm package.", could you please let me know how to install that? I try to run "llvm", seems the system doesn't know that either.

shrik3:
> @shrik3 for the "install the llvm package.", could you please let me know how to install that? I try to run "llvm", seems the system doesn't know that either.

I assume you are using ubuntu?

does `sudo apt install llvm` work for you? Perhaps you need to do `sudo apt update`first

PR #1270, opened 2024-05-21
aarch64: fix aarch64 syscalls after 425

CharlyYu:
Some syscalls after 425 should be implemented, or set to NoSys, otherwise any syscall from user mode will make the sandbox panic.


shrik3:
> Some syscalls after 425 should be implemented, or set to NoSys, otherwise any syscall from user mode will make the sandbox panic.

PR looks good to me. Do you have an example where the user calls an unsupported syscall?

shrik3 (approved):

CharlyYu:
Actually it is also the syscall 439 panic that makes me noticed that the syscalls after 425 is not handled properly. So I checked all  the syscalls  according to linux. @shrik3

shrik3:
you could merge this PR already, is there anything blocking?

PR #1274, opened 2024-05-22
epoll: define EpollEvent structure per linux API

### LINKED ISSUE #1230: aarch64: extremely slow dns look up.
will look into this later

shrik3:
the userData is sometimes defined as [i32; 2] in Quark code. This doesn't make sense to me because the data is used in one piece as a u64 anyways. But I may be wrong.

However I don't see where this array is converted into u64, am I missing something? What's the initial idea of define the userData as [i32; 2] ?

This is tested on both x86 and aarch64, I don't see regression so far. Also fixes #1230 

CC @QuarkContainer

### DIFF
--- a/qkernel/src/syscalls/sys_epoll.rs
+++ b/qkernel/src/syscalls/sys_epoll.rs
@@ -49,7 +49,7 @@ pub fn AddEpoll(
     fd: i32,
     flags: EntryFlags,
     mask: EventMask,
-    userData: [i32; 2],
+    userData: u64
 ) -> Result<()> {
     // Get epoll from the file descriptor.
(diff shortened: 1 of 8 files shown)


shrik3:
Note: @CharlyYu fixed a similar issue recently, but the EpollEvent struct is defined in more than 4 places differently depending on how they are used and they all require additional padding for aarch64. I don't want to have 8x struct definitions of the same thing so I propose to use one unified def of this struct.

The `Data` field could be either a fd (i32), a u32, or a u64 data. If you use `Data as i32` it will use the lower 4 bytes and has no conflict with the earlier defs.

PR #1288, opened 2024-05-24
[RFC] enable alternative installation and log path

shrik3:
**motivation**:  
(especially for development) we don't want to use `sudo` all the time. Also it doesn't make sense to hardcode the paths in the program.

**what's changing**:  
the makefile no longer uses `sudo` for the `install` target, you may need to run `sudo make install` if `make install` fails. Otherwise nothing changes if you don't modify the makefile.

**how to install to custom paths**:  
modify the makefile, and quark will use these paths for `qkernel`, `vdso.so` binary paths, config path and log path.

```
export PREFIX          ?= /home/shrik3/.local
export QBIN_DIR        ?= $(PREFIX)/bin
export QCONFIG_DIR     ?= /home/shrik3/.config/quark
export QLOG_DIR        ?= /tmp/log/quark
```
and modify the `/etc/docker/daemon.json` accordingly.

**limitations**  
1. quark doesn't create the `QLOG_DIR` (yet). We can do it in the makefile.  
2. other logs such as used by `qservice` are still created under the default path `/var/log/quark` but they can opt in pretty easily.

### DIFF
--- a/qvisor/src/main.rs
+++ b/qvisor/src/main.rs
@@ -61,6 +61,8 @@ extern crate regex;
 extern crate simplelog;
 extern crate spin;
 extern crate tabwriter;
+#[macro_use]
+extern crate const_format;
 
 #[macro_use]
 pub mod print;
@@ -168,8 +170,6 @@ lazy_static! {
         io_uring::IoUring::new(MemoryDef::QURING_SIZE as u32).expect("setup io_Uring fail")
     );
 }
-
-pub const LOG_FILE: &'static str = "/var/log/quark/quark.log";
 
 pub fn InitSingleton() {
     self::qlib::InitSingleton();
--- a/qvisor/Cargo.lock
+++ b/qvisor/Cargo.lock
@@ -204,6 +204,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "const_format"
+version = "0.2.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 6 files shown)


qvisor/src/main.rs
```diff
@@ -5,9 +5,7 @@
 // You may obtain a copy of the License at
 //
 //     http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
+// Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS,
```
shrik3, line 8: oops this is a mistake.

makefile
```diff
@@ -1,13 +1,14 @@
...
-PREFIX          ?= /usr/local
-QBIN_DIR        ?= $(PREFIX)/bin
-QCONFIG_DIR     ?= /etc/quark
-QLOG_DIR        ?= /var/log/quark
-
+export PREFIX          ?= /usr/local
+export QBIN_DIR        ?= $(PREFIX)/bin
+export QCONFIG_DIR     ?= /etc/quark
+export QLOG_DIR        ?= /var/log/quark
+export ARCH 		   ?= ${shell uname -m}
```
shrik3, line 8: shouldn't add tab here.

makefile
chl337, line 1: Could you please add a condition to check which files to copy over? It is annoying to get false negatives. I have a (semi) fix for that in the previous version, maybe do something similar.

Cheers

PR #1295, opened 2024-05-27
remove nvidia virtual kernel driver

QuarkContainer:


QuarkContainer:
In our new Nvidia virtualization system, we don't need to support Nvidia kernel driver virtualization. Remove that to clean the system.

chengchen666 (approved):

shrik3 (approved):

PR #1306, opened 2024-06-02
Hostalloc

QuarkContainer:


qlib/kernel/memmgr/mm.rs
```diff
@@ -422,7 +421,11 @@ impl MemoryManager {
         let (mut vseg, vgap) = mapping.vmas.Find(0);
         if vgap.Ok() {
             vseg = vgap.NextSeg();
-            debug!("VM: Start clean VMAs - KEY.0 ->\nvseg:{:?}\nvgap:{:?}", vseg.Value(), vgap.Range());
+            debug!(
```
shrik3, line 424: why "nv" here? (I suppose it's nvidia related?)

chengchen666: Yes, to implement one of cuda API

PR #1308, opened 2024-06-05
toolchain: enable using cargo's own build-std

shrik3:
we need to build the toolchain because qkernel (arch-qkernel.json) is not one of the default targets. We have been using the cargo-xbuild tool for this and it worked fine, however it's not  under active maintance for ~ 2 years and there could be some issues with newer rust toolchain as of May, 2024.

Rust's build-std is a unstable feature that does the same thing: we tell it to build the components e.g. core, alloc, builtins ... in the .cargo/config.toml so that cargo builds them with the custom target specified by `--target xyz.json`. To use this, simply replace the `cargo xbuild` with `cargo build` in qkernel's makefile, nothing else needs to change. (this approach is also noted in cargo-xbuild's docs)

Since cargo xbuild is not breaking (yet) and the build-std is not a unstable feature (yet), I'm keeping the cargo-xbuild as default. The build-std could opt-in if needed.

### DIFF
--- a/qkernel/.cargo/config.toml
+++ b/qkernel/.cargo/config.toml
@@ -1,0 +1,3 @@
+[unstable]
+build-std = ["core","alloc", "compiler_builtins"]
+build-std-features = ["compiler-builtins-mem"]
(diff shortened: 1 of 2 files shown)


shrik3:
the first commit simply adds a config file, it has no impact on the xbuild. in case something breaks, simply revert the second commit.

QuarkContainer (approved):
Tested. No issue found.

shrik3:
> Tested. No issue found.

thanks. I'll update the readme to remove the xbuild dependency later.

PR #1309, opened 2024-06-06
chore: URING: remove deadcode

chl337:


chl337:
@shrik3 ping:)

QuarkContainer (commented):
The code is related to the IORING_REGISTER_FILES_UPDATE feature. In some old Linux version (might be 5.8), we have to add register the fd before use them with IO_Uring. But if we enable support this, we have to create a large array which might cover all the fd range for example 0~1024 or 0~4096 etc. As some application might use very large amount fds, it will take much memory. 

I tried to add the configuraiton UringFixedFile to enable/disable this feature for Linux version compatibility. But I forget to update the code. 

Ideally, we'd better enable/disable this based on Linux version and expend and shrink the fd array.

I think we'd better to keep the code and add a issure or work item to enable the feature when we have bandwidth.

shrik3:
> I think we'd better to keep the code and add a issure or work item to enable the feature when we have bandwidth.

@chl337  In which case, please mark the functions in question with `todo!()`, comment out the references (with some notes perhaps) and write a memo somewhere (or maybe open an issue).

chl337:
> comment out the references (with some notes perhaps) 

That's a ghost state that I am not really fond with.

QuarkContainer:
@chl337 I agree with you. Please remove this now. Could you please also file an issue with the commit id so that we can reference the code when need it.

QuarkContainer (approved):

PR #1310, opened 2024-06-06
[Draft] cc rebase

123abcpp:
To enable cc mode, compiling with make cc_all/ cc_release/ cc_debug.
Set CCMode in config.json as **Normal** or **NormalEmu** . Normal uses identical mapping and NormalEmu uses unidentical mapping.


qlib/kernel/fs/host/util.rs
```diff
@@ -286,7 +293,10 @@ pub fn TryOpenAt(dirfd: i32, name: &str, skiprw: bool) -> Result<(i32, bool, Lib
         fstat: &fstat,
         writeable: false,
     };
+    #[cfg(not(feature = "cc"))]
     let cstr = CString::New(&name);
+    #[cfg(feature = "cc")]
+    let cstr = SharedString::New(&name);
```
shrik3, line 299: I see this pattern a lot. IMHO this doesn't scale well: whenever you need to share a string with the host you need to repeat this pattern, and those who doesn't work on cc would easily ignore it and breaks the cc code.

I would suggest using "SharedString" wherever it's used for guest/host communication. And implement them differently for cc/non-cc:
for cc: it's allocated on the shared heap, for non-cc, it's the same as CString.

Also in general, I think it should be explicit that an object is for sharing, be it cc or not.

qlib/kernel/socket/hostinet/tsotsocket.rs
```diff
@@ -405,9 +405,15 @@ impl TsotSocketOperations {
     }
 
     pub fn PostConnect(&self) {
+        #[cfg(not(feature = "cc"))]
```
shrik3, line 408: see my comments on the SharedString. The same argument applies.

qlib/kernel/quring/uring_mgr.rs
```diff
@@ -365,15 +387,22 @@ impl QUring {
         srcs: &[IoVec],
         ops: &TsotSocketOperations,
     ) -> Result<i64> {
-        let (count, writeBuf) = buf.Writev(task, srcs)?;
+        #[cfg(feature = "cc")]
```
shrik3, line 390: For a placeholder, I suggest leave this function as is (with `#[cfg(not(feature = "cc"))]`), make a cc version of it and make it `todo!()`

The same goes for other places where you have e.g. `panic!("not supported by cc");`. The idea is to make changes minimal, if the cc version is not ready.

qlib/kernel/quring/uring_async.rs
```diff
@@ -1098,14 +1181,20 @@ impl AIOWrite {
...
     }
 }
 
 pub struct AIOReadInner {
     pub fd: i32,
     pub buf: DataBuff,
+    #[cfg(not(feature = "cc"))]
     pub iovs: Vec<IoVec>,
+    #[cfg(feature = "cc")]
+    pub iovs: Vec<IoVec, GuestHostSharedAllocator>,
```
shrik3, line 1197: I don't know you could give Vec an allocator :D

shrik3 (commented):
overall, I'm seeing very little (if any) documentations. confidential computing is a heavy feature and introduces some breaking changes in the APIs. Please add documentations otherwise the affected code will become obscure to those who don't work in this area.

qlib/bytestream.rs
```diff
@@ -197,13 +197,23 @@ impl SocketBufIovs {
...
 }
 
+#[cfg(not(feature = "cc"))]
 pub enum RingeBufAllocator {
     HeapAllocator,
     //ShareAllocator {headTailAddr: u64, bufAddr: u64},
     ShareAllocator(u64, u64, u64, bool),
 }
 
+#[cfg(feature = "cc")]
```
shrik3, line 207: there is `bytestream_cc.rs`, why modify the original then?

qvisor/src/vmspace/mod.rs
```diff
@@ -159,6 +164,10 @@ impl VMSpace {
             vcpus: Vec::new(),
             haveMembarrierGlobal: haveMembarrierGlobal,
             haveMembarrierPrivateExpedited: haveMembarrierPrivateExpedited,
+            #[cfg(feature = "cc")]
```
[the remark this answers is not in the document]
123abcpp: Actually previously these members were not stored in VMSpace because they are used directly later in the same function as parameters of initSharespace, while now cc initiates sharespace later, so they are stored here. And if you think it is better to store them whether it is cc, it is an alternative.

qlib/kernel/Kernel.rs
shrik3, line 1: one reason to have `xyz_cc.rs` is that you don't have to modify the `xyz.rs`. Or are there other reasons to modify the original? Same goes for `bytestream.rs`

qlib/pagetable.rs
```diff
@@ -664,6 +690,93 @@ impl PageTables {
...
+
+        #[cfg(target_arch = "x86_64")]
+        let hugepage_flags = flags | PageTableFlags::HUGE_PAGE;
+
+        let mut res = false;
+
+        let mut curAddr = start;
+
+        let identical = IDENTICAL_MAPPING.load(Ordering::Acquire);
+        let pt: *mut PageTable = if identical{
```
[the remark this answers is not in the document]
123abcpp: Identical means identical mapping here, and it is a "cc" feature because unidentical mapping is enabled in CCMode NormalEmu, which is used to test whether cc modification works or not.

qlib/qmsg/sharepara.rs
shrik3, line 1: Perhaps there are better places for the sharepara module, since it's used by the hypercall, not directly by the qmsg.

qlib/kernel/taskMgr.rs
```diff
@@ -56,6 +66,8 @@ pub fn CreateTask(runFnAddr: u64, para: *const u8, kernel: bool) {
 extern "C" {
     pub fn context_swap(_fromCxt: u64, _toCtx: u64, _one: u64, _zero: u64);
     pub fn context_swap_to(_fromCxt: u64, _toCtx: u64, _one: u64, _zero: u64);
+    #[cfg(feature = "cc")]
+    pub fn context_swap_cc(_fromCxt: u64, _toCtx: u64, _one: u64, _zero: u64);
```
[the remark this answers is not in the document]
123abcpp: fixed

qlib/rdma_svc_cli.rs
```diff
@@ -438,6 +467,33 @@ impl RDMASvcClient {
                                         ),
                                     ));
 
+                                    #[cfg(feature = "cc")]
```
shrik3, line 470: well, this is kinda ugly but I don't know if there is better way ... :(

qlib/kernel/arch/x86_64/syscall_entry.s
```diff
@@ -152,6 +152,31 @@ context_swap:
...
+    mov [rdi+0x00], rsp
+    mov [rdi+0x08], r15
+    mov [rdi+0x10], r14
+    mov [rdi+0x18], r13
+    mov [rdi+0x20], r12
+    mov [rdi+0x28], rbx
+    mov [rdi+0x30], rbp
+
+    sfence
+    movq [rdx+0x00], 0x1
```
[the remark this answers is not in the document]
123abcpp: They both set 1 and 0 to the ready, maybe you mean something else?

qvisor/src/elf_loader.rs
```diff
@@ -166,6 +168,113 @@ impl KernelELF {
         return Ok(entry);
     }
 
+    #[cfg(feature = "cc")]
```
shrik3, line 171: how different is it to load the kernel with cc enabled? The repetition is pretty heavy here.

qvisor/src/runc/runtime/sandbox_process.rs
```diff
@@ -200,8 +200,8 @@ impl SandboxProcess {
         if taskSockFd > 0 {
             args.Pivot = false;
         }
-
-        let exitStatus = match VirtualMachine::Init(args) {
+        let cc_mode = QUARK_CONFIG.lock().CCMode;
```
shrik3, line 203: This would receive changes from @chl337 I suppose?

qvisor/src/runc/runtime/signal_handle.rs
```diff
@@ -115,18 +119,40 @@ extern "C" fn handle_sigintAct(
...
+                    SignalDeliveryMode::DeliverToProcess
+                },
+            };
+
+            VMS.lock().Signal(signal);
+        }
+
+        #[cfg(feature = "cc")]
+        if is_cc_enabled(){
+            let res = SignalProcess(&ROOT_CONTAINER_ID.lock().clone(), 0, signal, console);
```
shrik3, line 139: This could be optimized to reduce code repetition.

qlib/kernel/fs/host/util.rs
123abcpp: I agree to use "SharedString" wherever it's used for guest/host communication.  I can move the cfg here and add it into ShareString for cc/non-cc. Then ShareString will be a common mod but not cc specific.

shrik3 (commented):
another global comment is that there are some heavy code repetition. Some could be optimized.

qlib/kernel/quring/uring_mgr.rs
123abcpp: I see.

qlib/kernel/arch/x86_64/syscall_entry.s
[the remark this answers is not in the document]
shrik3: I mean why not setting it in the rust code upon context switch? Or do I not understand the purpose of this ready bit correctly?

qvisor/src/vmspace/mod.rs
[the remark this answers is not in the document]
shrik3: Perhaps just store them here for non-cc as well, I don't see a performance loss to store some additional data here.

qlib/pagetable.rs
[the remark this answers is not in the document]
shrik3: do you mean the unidentical mapping does not work for non-cc (yet)?

If non-identical mapping is also planed for non-cc, then this feature doesn't need to be cc only. If the identical-mapping config breaks non-cc when turned on, then let it be.

qlib/kernel/arch/x86_64/syscall_entry.s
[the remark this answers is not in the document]
123abcpp: I am not sure cause there are sfences between, may ask the maintainer if the ready bit setting can move forward.

qlib/bytestream.rs
123abcpp: Since there are repeated implementations.

qlib/qmsg/sharepara.rs
123abcpp: Yeah, I just want to put it near qcall, and you can absolutely move it somewhere else.  So where do you think is better?

qlib/kernel/fs/host/util.rs
shrik3: > Then ShareString will be a common mod but not cc specific.

It doesn't need to be CC specific. This is a general concept that "xyz" is for guest/host communication. Also regarding naming, CString and String means different things I think. CString is a null-terminated char array, while String is a more complicated object on the heap.

Anyways, I'm not saying you need to do the changes _now_ because we are looking at a systematic problem regarding host/guest comminication. sooner or later we need to deal with the interface. Passing a raw pointer to the host is very unsafe.

qlib/bytestream.rs
shrik3: Yes, but in the original you have modifications like
```
fn xyz() {
    if cc {
        do abc
    } else {
        do otherwise
}
```
Then it's technically two functions. Why not have a clean implementation of xyz in both files?

qvisor/src/elf_loader.rs
123abcpp: Not much, I can merge them.

qlib/bytestream.rs
123abcpp: In btyestream.rs? you mean fn Init()?

shrik3: sorry I mean kernel.rs, mostly.

qvisor/src/runc/runtime/signal_handle.rs
123abcpp: Sure, I can extract a function here.

qlib/pagetable.rs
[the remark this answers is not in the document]
123abcpp: I think currently this is used only for cc.

qlib/bytestream.rs
123abcpp: I will try to fix the Kernel.rs part.

qlib/mod.rs
```diff
@@ -1186,14 +1192,20 @@ impl CompleteEntry {
 }
 
 pub struct UringQueue {
+    #[cfg(not(feature = "cc"))]
     pub submitq: QMutex<VecDeque<UringEntry>>,
+    #[cfg(feature = "cc")]
+    pub submitq: ArrayQueue<UringEntry>,
```
shrik3, line 1198: why not Mutex in CC? And why ArrayQueue instead of VecDeque?

qlib/config.rs
```diff
@@ -138,4 +140,26 @@ pub enum CudaMemType {
...
+pub enum CCMode {
+    None,
+    Normal,
+    NormalEmu,
+    SevSnp,
+    Max
+}
+
+impl CCMode {
+    pub fn from(value: u64) -> Self {
```
shrik3, line 155: why mapping from numeric values? Aren't this enum user defined?

qlib/kernel/Kernel.rs
```diff
@@ -13,21 +13,40 @@
 // limitations under the License.
 
 use super::super::common::*;
+#[cfg(not(feature = "cc"))]
```
shrik3, line 16: well, this is pretty noisy, perhaps you could simply just import them, the current config won't warn unused imports anyways

qkernel/src/lib.rs
```diff
@@ -133,10 +145,23 @@ pub static VCPU_ALLOCATOR: GlobalVcpuAllocator = GlobalVcpuAllocator::New();
 pub static GLOBAL_ALLOCATOR: HostAllocator = HostAllocator::New();
 //pub static GLOBAL_ALLOCATOR: BitmapAllocatorWrapper = BitmapAllocatorWrapper::New();
 
+#[cfg(feature = "cc")]
+pub static  IS_GUEST: bool = true;
```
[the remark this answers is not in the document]
shrik3: oh I recall they should be const by default....

qlib/task_mgr.rs
```diff
@@ -212,9 +265,15 @@ pub struct TaskQueueIntern {
...
         return Self {
+            #[cfg(not(feature = "cc"))]
             workingTask: TaskId::New(0),
+            #[cfg(feature = "cc")]
+            workingTask: TaskId::New(0, 0),
             workingTaskReady: false,
+            #[cfg(not(feature = "cc"))]
             queue: VecDeque::with_capacity(8),
+            #[cfg(feature = "cc")]
+            queue: VecDeque::with_capacity(512),
```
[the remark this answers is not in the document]
123abcpp: Avoid runtime dynamic allocation.

qlib/kernel/threadmgr/task_exit.rs
```diff
@@ -912,7 +912,10 @@ impl Thread {
 
         let creds = self.Credentials();
         let kuid = creds.lock().RealKUID;
+        #[cfg(not(feature = "cc"))]
         let userns = receiver.UserNamespace();
+        #[cfg(feature = "cc")]
+        let userns = creds.lock().UserNamespace.clone();
```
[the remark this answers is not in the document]
123abcpp: I do not call this function, I use the attribute directly.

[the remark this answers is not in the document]
shrik3: > I do not call this function

hmmm I missed the `()` .. but anyways why not just call it?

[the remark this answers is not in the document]
123abcpp: > > I do not call this function
> 
> hmmm I missed the `()` .. but anyways why not just call it?
As the commit log, to avoid deadlock. Actually I don't know why the deadlock happened, but the lock() inside UserNamespace() causes a deadlock.

qlib/qmsg/sharepara.rs
```diff
@@ -0,0 +1,26 @@
...
+    pub para3: u64,
+    pub para4: u64,
+}
+
+pub const SHAREPARA_COUNT: u64 = MemoryDef::PAGE_SIZE/core::mem::size_of::<SharePara>() as u64;
+
+#[derive(Debug, Copy, Clone)]
+#[repr(C)]
+pub struct ShareParaPage {
+    pub SharePara: [SharePara ;SHAREPARA_COUNT as usize],
```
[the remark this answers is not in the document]
123abcpp: > I don't get this. Why does it have to be an array where you only use the first element?
> 
> 1. `ShareParaPage.SharePara[0]` is a wrapper of `para1,para2,para3,para4` , there is a semantic gap
> 2. this allows accessing undefined fields (SharePara[1] onwards), while doing no harm, this is undefined and should be prevented.
> 3. initialize this array is is .... kinda expensive? here you have a 4k page, where the 4 parameters are 32 bytes. Meaning you have to zero out ~4K bytes every time you create one such thing

ShareParaPage is a page with fix address, and with each vcpu, it has its own SharePara entry according to its vcpu id to do the host guest communication. This struct will not be initialized, but just use the address and then pass the parameters.

qlib/socket_buf.rs
```diff
@@ -44,9 +50,34 @@ impl SocketBuffWeak {
...
+
+#[cfg(feature = "cc")]
+impl Default for SocketBuffWeak{
+    fn default() -> Self {
+        SocketBuff::default().Downgrade()
+    }
+}
+
+#[cfg(feature = "cc")]
+impl SocketBuffWeak {
```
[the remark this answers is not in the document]
123abcpp: For the upgrade function yes, I will merge this function.

qlib/config.rs
123abcpp: Used for reusing the sharedspace u64 passed to the kernel as the CCMode when cc is enabled.

qlib/mod.rs
123abcpp: 1.ArrayQueue is from Crossbeam crate, which is designed for lock free concurrency.
2.ArrayQueue instead of VecDeque: avoid dynamic growth, arrayqueue will return an error when the queue is full.

qlib/kernel/Kernel.rs
123abcpp: It will warn unused imports...

PR #1319, opened 2024-06-17
remove uring dependency for controlsock process

QuarkContainer:


qlib/fileinfo.rs
```diff
@@ -241,6 +252,27 @@ impl FdWaitInfo {
...
+
+            let mask = fi.queue.Events();
+            if mask == fi.mask {
+                return Ok(());
+            }
+
+            mask
+        };
+
+        return Self::waitfd(fd, mask);
```
chl337, line 267: Can't the body of the function be called directly here?

qlib/kernel/quring/uring_mgr.rs
```diff
@@ -556,24 +552,9 @@ impl QUring {
...
-            Some(cqe) => {
-                self.Process(&cqe);
-                return true;
-            }
-        }
-    }
-
     pub fn DrainCompletionQueue(&self) -> usize {
         if SHARESPACE.config.read().UringIO {
+            error!("DrainCompletionQueue xxxx");
```
chl337, line 557: Is it indeed an error if the path is taken? What does the "xxxx" mean?

qvisor/src/vmspace/host_uring.rs
```diff
@@ -60,61 +58,59 @@ pub fn CopyCompleteEntry() -> usize {
...
-        }
-        assert!(sq.dropped() == 0, "dropped {}", sq.dropped());
-        assert!(!sq.cq_overflow());
-
-        while !sq.is_full() {
-            let uringEntry = match submitq.pop_front() {
-                None => break,
-                Some(e) => e,
-            };
+            if sq.dropped() != 0 {
```
chl337, line 71: This mixture of if-s and assert-s seems somewhat confusing to me.

qlib/kernel/quring/uring_mgr.rs
QuarkContainer: It is by mistake. Forget to remove the debug comments.

qvisor/src/vmspace/host_uring.rs
QuarkContainer: Remove if-s/

qlib/fileinfo.rs
QuarkContainer: sorry that I don't understand the question. where can the function body called directly?

chl337: What I meant is, why not calling
```
WaitFDAsync(...)
```
directly?

qlib/linux_def.rs
```diff
@@ -2256,6 +2259,22 @@ impl PermMask {
...
+#[cfg(target_arch = "x86_64")]
+impl Flags {
+    pub const O_DIRECT: i32 = 0o00040000; //0x00004000;
+    pub const O_LARGEFILE: i32 = 0o00100000; //0x00008000;
+    pub const O_DIRECTORY: i32 = 0o00200000; //0x00010000;
+    pub const O_NOFOLLOW: i32 = 0o00400000; //0x00020000;
+}
+
+#[cfg(target_arch = "aarch64")]
+impl Flags {
```
chl337, line 2271: Maybe open a new PR for this so that we can merge it already?
@mhomidi Do you want to add your signoff on it?!

qlib/fileinfo.rs
QuarkContainer: I still doesn't understand. WaitFDAsync needs call UpdateFDSync. If we can WaitFDAsync in UpdateFDSync, it is recursive call.

chl337: I don't know if we are talking about the same thing :). What I see here is the call-chain:
```
fn UpdateFDSync(...)
       |
      ...
       |
  Self::waitfd(...)   ------>  fn waitfd(x,y)
       |                          |  
       |                      HostSpace::WaitFDAsync(x,y) --------> ....
       |                          |
       |             <------    Ok(()) 
      ///
```
Because `fn waitfd(...)` does not apply any logic to the arguments, I would guess that it is equivalent to:
```
fn UpdateFDSync(...)
       |
      ...
       |
  HostSpace::WaitFDAsync(x,y) --------> ....
       | 
    Ok(())
       |
      ///
```
except from the additional stack layout that gets created by the function call. That is why I asked.

* [fn waitfd(...)](https://github.com/QuarkContainer/Quark/blob/e75568ce4123e9ed854d09ef284ccfbc46fb3754/qlib/fileinfo.rs#L270)
* [fn WaitFDAsync(...)](https://github.com/QuarkContainer/Quark/blob/e75568ce4123e9ed854d09ef284ccfbc46fb3754/qlib/kernel/Kernel.rs#L937)

mhomidi (approved):

PR #1322, opened 2024-06-21
[RFC]Refactor creation and execution of VM and vCPU

chl337:


qvisor/src/arch/aarch64/vcpu/kvm_vcpu.rs
```diff
@@ -0,0 +1,162 @@
...
+pub const _SCTLR_UCT        :u64 = 1 << 15;
+pub const _SCTLR_SPAN       :u64 = 1 << 23;
+pub const _SCTLR_UCI        :u64 = 1 << 26;
+pub const _SCTLR_EL1_DEFAULT:u64 = _SCTLR_M | _SCTLR_C | _SCTLR_I | _SCTLR_UCT
+                                    | _SCTLR_UCI | _SCTLR_DZE | _SCTLR_SPAN;
+
+#[repr(u64)]
+#[derive(Copy, Clone)]
+pub enum KvmAarch64Reg {
+    R0     = 0x6030000000100000,
```
shrik3, line 91: why "R" instead of "X"?

qvisor/src/arch/tee.rs
shrik3, line 1: tee as in trusted execution env?

qvisor/src/runc/runtime/vm_type.rs
```diff
@@ -0,0 +1,42 @@
...
+
+pub mod noncc;
+pub mod resources;
+
+use std::sync::Arc;
+use kvm_ioctls::{Kvm, VmFd};
+use crate::{arch::vm::vcpu::ArchVirtCpu, elf_loader::KernelELF, qlib::common::Error, runc::runtime};
+use runtime::{vm::VirtualMachine, loader::Args};
+
+pub trait VmType: std::fmt::Debug {
```
shrik3, line 24: I don't see why a supertrait is needed here, this trait doesn't provide any default impl that uses the Debug trait.

qvisor/src/runc/runtime/vm_type.rs
shrik3, line 1: why `vm_type.rs` is under `runtime` while there is another `vm_type` dir? shouldn't this file be moved to `vm_type/mod.rs` ?

chl337:
On Fri Jun 21, 2024 at 3:13 PM CEST, Tianhao Wang wrote:
> @shrik3 commented on this pull request.
>
>
>
> > @@ -277,7 +277,10 @@ impl PageOpts {
>  
>      pub fn Kernel() -> Self {
>          return PageOpts(
> -            PageTableFlags::VALID,
> +            PageTableFlags::VALID |
> +            PageTableFlags::MT_NORMAL |
> +            PageTableFlags::DIRTY |
>
> why should this template PTE flag have dirty and accessed bit set? shouldn't that be done explicitly?
A shameful CP, let me check if they are redundant.

> > +pub const _CNTKCTL_EL1_DEFAULT:u64 = _CNTKCTL_EL0PCTEN | _CNTKCTL_EL0VCTEN;
> +
> +pub const _SCTLR_M          :u64 = 1 << 0;
> +pub const _SCTLR_C          :u64 = 1 << 2;
> +pub const _SCTLR_I          :u64 = 1 << 12;
> +pub const _SCTLR_DZE        :u64 = 1 << 14;
> +pub const _SCTLR_UCT        :u64 = 1 << 15;
> +pub const _SCTLR_SPAN       :u64 = 1 << 23;
> +pub const _SCTLR_UCI        :u64 = 1 << 26;
> +pub const _SCTLR_EL1_DEFAULT:u64 = _SCTLR_M | _SCTLR_C | _SCTLR_I | _SCTLR_UCT
> +                                    | _SCTLR_UCI | _SCTLR_DZE | _SCTLR_SPAN;
> +
> +#[repr(u64)]
> +#[derive(Copy, Clone)]
> +pub enum KvmAarch64Reg {
> +    R0     = 0x6030000000100000,
>
> why "R" instead of "X"?
I was thinking "R" as "Register", but "X" is probably better... I'll
change it.
>
> On qvisor/src/arch/tee.rs:
>
> tee as in trusted execution env?
Ja, although naming may change/removed in further revision - see, my
first concern is having a general template first, so something like this
can be seen as a placeholder in some sort of away... Apologies for my
rookieness... 
> > +// Unless required by applicable law or agreed to in writing, software
> +// distributed under the License is distributed on an "AS IS" BASIS,x
> +// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
> +// See the License for the specific language governing permissions and
> +// limitations under the License.
> +
> +
> +pub mod noncc;
> +pub mod resources;
> +
> +use std::sync::Arc;
> +use kvm_ioctls::{Kvm, VmFd};
> +use crate::{arch::vm::vcpu::ArchVirtCpu, elf_loader::KernelELF, qlib::common::Error, runc::runtime};
> +use runtime::{vm::VirtualMachine, loader::Args};
> +
> +pub trait VmType: std::fmt::Debug {
>
> I don't see why a supertrait is needed here, this trait doesn't provide any default impl that uses the Debug trait.
No, it is indeed needed.
> On qvisor/src/runc/runtime/vm_type.rs:
>
> why `vm_type.rs` is under `runtime` while there is another `vm_type` dir? shouldn't this file be moved to `vm_type/mod.rs` ?
```
- foo.rs foo/
         |_whatever.rs
```
is the recommended way (since 2018 I think), and I kind of like it:)

Cheers

chl337:
@123abcpp ping

123abcpp:
Just a reminder here, as we have discussed before, we need to initialize the private and shared allocator somewhere(at the end of init function or add a new function for trait vm_type). Then just mmap a small range of area for the initial allocator before these allocators are initialized.

shrik3:
>> I don't see why a supertrait is needed here, this trait doesn't provide any default impl that uses the Debug trait.
>
> No, it is indeed needed.

whatever requires the Debug trait, is not within the trait def. I guess you have a function somewhere that takes a VM struct that implements `VmType` and prints info about it. But that's external to this trait def.

qvisor/src/arch/aarch64/vcpu/kvm_vcpu.rs
```diff
@@ -0,0 +1,162 @@
...
+pub const _MT_NORMAL            :u64 = 4;
+pub const _MT_NORMAL_WT         :u64 = 5;
+pub const _MT_ATTR_DEVICE_nGnRnE:u64 = 0x00;
+pub const _MT_ATTR_DEVICE_nGnRE :u64 = 0x04;
+pub const _MT_ATTR_DEVICE_GRE   :u64 = 0x0c;
+pub const _MT_ATTR_NORMAL_NC    :u64 = 0x44;
+pub const _MT_ATTR_NORMAL_WT    :u64 = 0xbb;
+pub const _MT_ATTR_NORMAL       :u64 = 0xff;
+pub const _MT_ATTR_MASK         :u64 = 0xff;
+pub const _MT_EL1_INIT: u64 = (_MT_ATTR_DEVICE_nGnRnE << (_MT_DEVICE_nGnRnE * 8))
```
shrik3, line 67: MAIR_EL1

qvisor/src/arch/aarch64/vcpu/kvm_vcpu.rs
```diff
@@ -0,0 +1,162 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#![allow(non_upper_case_globals)]
+use kvm_bindings::kvm_vcpu_events;
+use crate::{KVMVcpu, qlib::common::Error, vmspace::VMSpace, qlib::backtracer};
+
+pub const _KVM_ARM_VCPU_PSCI_0_2: u32 = 2;
```
shrik3, line 19: btw ... why did we name this register encoding with a leading underscore in the first place, when they are `pub` ?

qvisor/src/arch/mod.rs
```diff
@@ -12,4 +12,48 @@
...
+    fn get_hypercall_arguments(&self, vcpu_fd: &VcpuFd, vcpu_id: usize)
+        -> Result<(u64, u64, u64, u64), Error>;
+    fn should_handle_kvm_exit(&self, kvm_exit: &VcpuExit) -> bool;
+    fn should_handle_hypercall(&self, hypercall: u16) -> bool;
+    fn handle_kvm_exit(&self, kvm_exit: &VcpuExit, vcpu_id: usize) -> Result<bool, Error>;
+    fn handle_hypercall(&self, hypercall: u16, data: &[u8], arg0: u64, arg1: u64, arg2: u64,
+        arg3: u64, vcpu_id: usize) -> Result<bool , Error>;
+}
+
+pub enum ConfCompType {
```
shrik3, line 57: what's this for? whether something needs a configuration?

shrik3: and there are `ConfCompType::NoConf`, and `tee::NonConf`, I'm already confused

shrik3: `Comp` as in?

qvisor/src/arch/aarch64/vcpu/kvm_vcpu.rs
shrik3: If I understand correctly a name prefixed with `_` means "this is not meant to be called/used", or "this is only used internally. In the later case these items should not be `pub` either. This feels like a relic from early development. It's a chance to fix them here. (same may goes for some other functions).

qvisor/src/runc/runtime/vm_type/noncc.rs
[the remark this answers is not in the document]
chl337: I agree and very open to suggestions. In internal conversations, it was referred to as "basic", "normal" or "non-secure", but IMO those terms are somewhat not appropriate, each for its own reasons. Here the name is influenced by what feature I deem has more influence over others/future-ones, so I am clearly biased.

PR #1326, opened 2024-07-12
fix redis signal handle failure bug (fixed by mhomidi Hadi Omidi)

QuarkContainer:


qlib/kernel/threadmgr/task_signals.rs
```diff
@@ -1310,7 +1310,6 @@ impl Task {
         let signo = info.Signo as u64;
         let currTask = Task::Current();
         let regs = currTask.GetPtRegs();
-        *regs = PtRegs::default();
```
shrik3, line 1313: that means, we are inheriting saved registers from the last interrupt, instead of setting them up from scratch? do you have idea what specific registers are necessary here?

QuarkContainer: I have no idea. This is also the behavior of X86 code and I copied from somewhere before. Looks golang signal process relies on that.

shrik3: I think reusing the whole trap frame may be a foot gun. Time to read the friendly manual!

For now, I think we may keep this PR open for a while as a temporary fix.

mhomidi: Probably we just need to keep the register that store the address of g and m structure in the memory in go-runtime. BTW, I agree with @shrik3 and we can keep this PR open for temporary fix.

chl337: @shrik3 If I understand it correctly, for arm64, go initially saves _g_ (whatever that nonsense name means) in [R12](https://github.com/golang/go/blob/90c6558b6acef5a9b9fb8f3c35cff58423c8b00e/src/runtime/sys_linux_arm64.s#L539). So you don't want to lose that one.

P.S. It may also move it again, before calling [sigtrampgo](https://github.com/golang/go/blob/90c6558b6acef5a9b9fb8f3c35cff58423c8b00e/src/runtime/signal_unix.go#L431), but I haven't checked it...

chl337:
@QuarkContainer Please change the commit message to something more informative, e.g.
> aarch64: preserve the GPR state - Linux ABI

otherwise it's O.K.

shrik3:
I dig it, we should not "reset" the pt_regs on signal delivery. The kernel code doesn't do that. I don't know why it is the case for x86.

I'll go ahead and merge this.

shrik3 (approved):

Issue #1330, opened 2024-07-18
Issue with gvisor tests when running on ARM


mhomidi:
Hi,

I tried to build the binary files of test of gvisor to run them on the Quark when our system is arm. The instruction is that you just need to follow [this](https://github.com/google/gvisor/tree/master/test/syscalls) instruction to build binary files of each test you want. For instance I tried to run for write syscall. So I just run
```
bazel test //test/syscalls:write_test_native
```
Then, by `bazel info bazel-bin` you can get the directory of binary files which is something like `/home/[USERNAME]/.cache/bazel/_bazel_[USERNAME]/c8d627c8e0d3a3fd49664c9d76222cd6/execroot/__main__/bazel-out/aarch64-fastbuild/bin`
Then, you just need to run `./test/syscalls/write_test_native`

Now if you do these steps on x86 and run on Quark, everything is okay and the output is:
```
+ [[ -n /htmp ]]
+ mkdir -p /htmp
+ chmod a+rwx /htmp
+ exec test/runner/runner_/runner --platform=native --platform-support=32BIT:TRUE,ALIGNMENT_CHECK:TRUE,INT3:TRUE,MULTIPROCESS:TRUE,VSYSCALL:TRUE --network=none --use-tmpfs=False --fusefs=False --file-access=exclusive --overlay=False --add-host-uds=False --add-host-connector=False --add-host-fifo=False --strace=True --debug=True --container=False --one-sandbox=True --iouring=False --directfs=False --leak-check=False --save=False --save-resume=False test/syscalls/linux/write_test
I0718 21:09:22.197586       1 main.go:170] Executing: [/gtest/test/syscalls/linux/write_test --gtest_filter=WriteTest.WriteNoExceedsRLimit]
Note: Google Test filter = WriteTest.WriteNoExceedsRLimit
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from WriteTest
[ RUN      ] WriteTest.WriteNoExceedsRLimit
[       OK ] WriteTest.WriteNoExceedsRLimit (1 ms)
[----------] 1 test from WriteTest (1 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (1 ms total)
[  PASSED  ] 1 test.
Failed to match any benchmarks against regex: .
PASS
```

But when you do it on an ARM processor, the output will be some thing like this:
```
+ [[ -n /htmp ]]
+ mkdir -p /htmp
[...]

PR #1331, opened 2024-07-25
aarch64: adjust UContext padding (sigset_t) to 128 Bytes

chl337:
Related to #1330.
cc: @mhomidi

### DIFF
--- a/qlib/kernel/SignalDef.rs
+++ b/qlib/kernel/SignalDef.rs
@@ -378,7 +378,8 @@ pub struct UContext {
     pub Link: u64,
     pub Stack: SignalStack,
     pub Sigset: u64,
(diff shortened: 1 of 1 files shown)


mhomidi:
I think we also need to add a 4-byte (u32) padding in [signalStack](https://github.com/QuarkContainer/Quark/blob/fd7555a73187c79b37eb96e97c526a8c692d2d26/qlib/kernel/SignalDef.rs#L911) after flags.

shrik3:
please take it to #1348 .

There are two missing padding: the sigstack fields should be padded to 8 bytes aligned and the sigcontext should be padded to 16 bytes aligned.

chl337:
Had almost forgotten about this....

chl337:
Had almost forgotten about this...

Taken to #1348

Issue #1337, opened 2024-08-01
Problem in context swapping in ARM


mhomidi:
Hi,

We have a problem in context switching in the ARM. Based on what we have in x86, when a task is running on a vcpu, `ready` in its context should be 0 but when we read the code of context swap in `qlib/kernel/arch/aarch64/syscall_entry.s:60`, we are storing 0 (x3 which is arg 4 in `context_swap` in `qkernel/src/qlib/kernel/taskMgr.rs:58`) in ready for task which is swapped out and store 1 for task swapped in. 
```asm

context_swap:
    mov     x8, x0
    mov     x9, sp
    isb
    stp     x19, x20, [x8], #16
    stp     x21, x22, [x8], #16
    stp     x23, x24, [x8], #16
    stp     x25, x26, [x8], #16
    stp     x27, x28, [x8], #16
    stp     x29, x9, [x8], #16
    str     lr, [x8], #16
    dsb ish
    str     x3, [x8] // Should be x2
    mov     x8, x1
    isb
    ldp     x19, x20, [x8], #16
    ldp     x21, x22, [x8], #16
    ldp     x23, x24, [x8], #16
    ldp     x25, x26, [x8], #16
    ldp     x27, x28, [x8], #16
    ldp     x29, x9, [x8], #16
    ldr     lr, [x8], #8
    ldr     x0, [x8], #8
    dsb ish
    mov     sp, x9
    str     x2, [x8] // Should be x3
    ret

```
It causes a problem when a vcpu want to steal a job from other vcpu's ready queue in `qkernel/src/qlib/task_mgr.rs:297`.
```rust
    // try to steal task from other vcpu's queue
    pub fn Steal(&self) -> Option<TaskId> {
        if self.queueSize.load(Ordering::Acquire) == 0 {
            return None;
        }

        match self.data.try_lock() {
            None => return None,
            Some(mut data) => {
                for _ in 0..data.queue.len() {
                    match data.queue.pop_front() {
                        None => panic!("TaskQueue none task"),
                        Some(taskId) => {
                            if taskId.GetTask().Ready() != 0 {
                                self.queueSize.fetch_sub(1, Ordering::Release);
                                return Some(taskId);
                            }
                            data.queue.push_back(taskId)
                        }
                    }
                }
            }
        }

        return None;
    }
```
The `ready` should be 1 but based on code of ctx swap in ARM, the task with this `ready` is running in another vcpu.

chl337:
Thanks for noticing!:)

PR #1338, opened 2024-08-03
aarch64: context_swap - adjust 3rd & 4rth parameter

chl337:
#1337 
P.S. @mhomidi

### DIFF
--- a/qlib/kernel/arch/aarch64/syscall_entry.s
+++ b/qlib/kernel/arch/aarch64/syscall_entry.s
@@ -69,7 +69,7 @@ context_swap:
     stp     x29, x9, [x8], #16
     str     lr, [x8], #16
     dsb ish
-    str     x3, [x8]
+    str     x2, [x8]
     mov     x8, x1
     isb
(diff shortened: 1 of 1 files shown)


shrik3:
btw I don't see why the context_swap abi has to provide 1 and 0 through parameters. Can't we just use asm literals?

chl337:
> btw I don't see why the context_swap abi has to provide 1 and 0 through parameters. Can't we just use asm literals?

I don't know if @QuarkContainer has a plan for 
 ``` rust
    pub fn context_swap(_fromCxt: u64, _toCtx: u64, _one: u64, _zero: u64);
```
Otherwise, I could drop the two last arguments.

QuarkContainer:
@chl337 I have no plan for that. Please go ahead to drop those.

mhomidi (approved):

QuarkContainer:
@chl337 is that ok to merge this before your new fix?

chl337:
I guess so... Unfortunately, I will not be active till 16/17.08

Cheers

PR #1348, opened 2024-08-26
Fix paddings in signal structs

### LINKED ISSUE #1330: Issue with gvisor tests when running on ARM

shrik3:
should fix #1330

supersedes #1331 (the second commit is technically doing the same as 1331, but 1331 was a wrong take. The sigset should be padded to 128 bytes (i.e. adding 120 bytes padding), the problem is that the sigcontext should be padded to 16 bytes.

There is also a alignment bug in the signalstack

test wanted.

@chl337 @mhomidi

### DIFF
--- a/qlib/kernel/SignalDef.rs
+++ b/qlib/kernel/SignalDef.rs
@@ -378,6 +378,9 @@ pub struct UContext {
     pub Stack: SignalStack,
     pub Sigset: u64,
     __pad: [u8; (1024 / 8) - core::mem::size_of::<u64>()],
+    __pad2: [u8; 8],
+    // MContext is 16 bytes aligned.
+    // offset here: 8 + 8 + 24 + 8 + 120 + 8 = 176
     pub MContext: SigContext,
 }
 
@@ -390,6 +393,7 @@ impl Default for UContext {
             Stack: SignalStack::default(),
             Sigset: 0,
             __pad: [0; (1024 / 8) - core::mem::size_of::<u64>()],
+            __pad2: [0; 8],
             MContext: SigContext::default(),
         }
     }
@@ -404,6 +408,7 @@ impl UContext {
             Stack: alt.clone(),
             Sigset: 0,
             __pad: [0u8; (1024 / 8) - core::mem::size_of::<u64>()],
+            __pad2: [0u8; 8],
             MContext: SigContext::New(ptRegs, oldMask, fault_address, fpstate),
         };
     }
@@ -905,6 +910,8 @@ pub struct SignalStruct {
 }
 
 // https://elixir.bootlin.com/linux/latest/source/arch/x86/include/uapi/asm/signal.h#L132
+// NOTE: repr(C) fields are self-aligned, i.e. the `size` is aligned to 8 bytes
+// and the SignalStack is in total 24 bytes instead of 20.
 #[derive(Debug, Clone, Copy)]
 #[repr(C)]
 pub struct SignalStack {


chl337:
b182c089f2671aa0d16e747e02fa157be2d26d0d has skipped below the radar for quite some time:)

>test wanted
I'll try to do it later.

shrik3:
btw, the linux source code doesn't specify the paddings explicitly, but you can certainly verify it with this cursed code:


```c
#include <signal.h>
#include <ucontext.h>
#include <stdio.h>

int main() {
	stack_t s;
	ucontext_t uc;
	printf("size of stack_t is %ld\n", sizeof(stack_t));
	printf("%ld %ld %ld\n", sizeof(s.ss_sp), sizeof(s.ss_flags), sizeof(s.ss_size));
	printf("%p %p %p\n", &s.ss_sp, &s.ss_flags, &s.ss_size );
	printf("size of ucontext %ld \n", sizeof(ucontext_t));
	printf("size of sigset %ld \n", sizeof(sigset_t));

	printf("%ld\n", (long)&(uc.uc_flags) - (long)&(uc));
	printf("%ld\n", (long)&(uc.uc_link) - (long)&(uc));
	printf("%ld\n", (long)&(uc.uc_stack) - (long)&(uc));
	printf("%ld\n", (long)&(uc.uc_sigmask) - (long)&(uc));
	printf("%ld\n", (long)&(uc.uc_mcontext) - (long)&(uc));
	return 0;
}
```

qlib/kernel/task.rs
```diff
@@ -903,10 +903,8 @@ impl Task {
         }
 
         if !alt.IsEnable() {
-            self.signalStack = SignalStack {
-                flags: SignalStack::FLAG_DISABLE,
-                ..Default::default()
-            }
+            self.signalStack = SignalStack::default();
```
chl337, line 906: Is this not äqv. to the previous implementation?

shrik3:
On Mon Aug 26, 2024 at 1:36 PM CEST, chl337 wrote:
> @chl337 commented on this pull request.
>
>
>
> > @@ -903,10 +903,8 @@ impl Task {
>          }
>  
>          if !alt.IsEnable() {
> -            self.signalStack = SignalStack {
> -                flags: SignalStack::FLAG_DISABLE,
> -                ..Default::default()
> -            }
> +            self.signalStack = SignalStack::default();
>
> Is this not äqv. to the previous implementation?

This should be equivalent.

because the `pad` is a private field; Using `..Default::default()` on private
fields will fail to compile.

-- 
wth
220F 6A50 7BEF 8C33 3B4A 4404 99DE B3B4 1583 4AEF

chl337:
> because the `pad` is a private field; Using `..Default::default()` on private fields will fail to compile.

I didn't know that, thanks.

shrik3:
b182c089f2671aa0d16e747e02fa157be2d26d0d is not necessary:

since the SignalStack struct is `#[repr(C)]`, it will be aligned to 24 bytes anyways. We may just drop the first commit.

qlib/kernel/SignalDef.rs
```diff
@@ -378,6 +378,9 @@ pub struct UContext {
     pub Stack: SignalStack,
     pub Sigset: u64,
     __pad: [u8; (1024 / 8) - core::mem::size_of::<u64>()],
+    __pad2: [u8; 8],
```
chl337, line 381: It should also be initialized in [1)](https://github.com/QuarkContainer/Quark/blob/22456041cd276b4143bfb63b2fe4d3e38bfabaa6/qlib/kernel/SignalDef.rs#L389) and [2)](https://github.com/QuarkContainer/Quark/blob/22456041cd276b4143bfb63b2fe4d3e38bfabaa6/qlib/kernel/SignalDef.rs#L403).

shrik3: Thanks, I'll update it.

chl337 (approved):
Works for me.

PR #1353, opened 2024-11-13
'Performance Test Result and Comparison' as a title

SherlockShemol:
I think 'Performance Test Result and Comparison' should be a title and it may well be the doc writer's true intention.

### DIFF
--- a/doc/perf_test.md
+++ b/doc/perf_test.md
@@ -1,4 +1,4 @@
-#Performance Test Result and Comparison
+# Performance Test Result and Comparison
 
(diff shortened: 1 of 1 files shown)


shrik3:
Thanks for the patch.

For a better git practice, I suggest renaming your commit. See [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for some inspiration.

For this commit specifically, you don't need to repeat verbatim what you fixed in the commit short log. I would put some thing simply like `docs: fix markdown syntax`.

shrik3 (approved):

shrik3:
> Thank you for your strict standards.Modified.

it will save you a lot of sanity in the future :) merged.

PR #1357, opened 2025-01-13
cc base support

123abcpp:
# Add a common base for feature Confidential Computing support:

  1. Enable using the heap with private (guest memory) and shared allocation.
  2. Enable creating VM based on the configurable features.
    - ATM: on CC modes

For the implementation details please refer to [Quark_CC.pdf](https://github.com/user-attachments/files/18399204/Quark_CC.pdf)

**Draft merge commit message**

```
This PR adds a common base for feature Confidential Computing support.

This PR is contributed by the Chair of Systems Engineering at TU Dresden as part of the series 
for adding Confidential Computing to the Quark framework.

Signed-off-by: Christo Bita <christo.bita@mailbox.tu-dresden.de>
Signed-off-by: yaoxin <yaoxin19950517@gmail.com>
Signed-off-by: Yiliang Dong <dongyiliangsteven@163.com>
Signed-off-by: Pamenas Kariuki <firstname.lastname@tu-dresden.de>
```


qlib/kernel/taskMgr.rs
```diff
@@ -55,7 +58,8 @@ pub fn CreateTask(runFnAddr: u64, para: *const u8, kernel: bool) {
 }
 
 extern "C" {
-    pub fn context_swap(_fromCxt: u64, _toCtx: u64);
+    pub fn context_swap(_fromCxt: u64, _toCtx: u64, _one: u64, _zero: u64);
```
shrik3, line 61: I thought this one/zero parameter hack was removed in earlier patches, why is it back?

qlib/kernel/task.rs
```diff
@@ -167,9 +183,45 @@ impl Drop for Task {
     }
 }
 
+// task wrapper on shared memory
+#[repr(C)]
+pub struct TaskWrapper {
```
shrik3, line 188: this is still necessary?

qlib/kernel/taskMgr.rs
123abcpp: Sorry, there is something wrong when rebasing this file, I will remove this.

qlib/kernel/task.rs
123abcpp: Yes, this is still necessary. When cc mode is on, the address of TaskWrapper will be stored as taskid instead of the address of task. We can get rid of further modifications but we still need to share the ready, queueid, and task address to host for scheduling.

qkernel/src/kernel_def.rs
```diff
@@ -269,42 +339,82 @@ pub fn NewSocket(fd: i32) -> i64 {
...
-        super::SHARESPACE.AQCall(&om);
-        taskMgr::Wait();
-        return qMsg.ret;
+            super::SHARESPACE.AQCall(&om);
+            taskMgr::Wait();
+            return qMsg.ret;
+        }
     }
 
     pub fn HCall(msg: &mut Msg, lock: bool) -> u64 {
```
shrik3, line 390: global comment: on frequently-called functions like, alloc, ctx switch, qcalls/hcalls/hypercalls, think about overheads of checking CC conditions.

qkernel/src/syscalls/sys_time.rs
```diff
@@ -14,6 +14,7 @@
 
 use alloc::boxed::Box;
 
+use crate::GUEST_HOST_SHARED_ALLOCATOR;
```
shrik3, line 17: perhaps unrelated:

the syscall gettimeofday is typically done via vdso on x86 systems. How does this (among other vsyscalls/vdso functions) work under CC context?

qkernel/src/kernel_def.rs
123abcpp: We have checked the performance of the batch of tests in https://github.com/QuarkContainer/Quark/blob/main/doc/perf_test.md, there is no performance loss compared with the current main branch Quark.

qlib/kernel/Kernel.rs
```diff
@@ -11,16 +11,24 @@
...
 use super::super::qmsg::*;
 use super::super::socket_buf::*;
 use super::super::*;
 use crate::kernel_def::HyperCall64;
+use crate::qlib::kernel::arch::tee::is_cc_active;
 use crate::qlib::proxy::*;
+use crate::GLOBAL_ALLOCATOR;
+use crate::GUEST_HOST_SHARED_ALLOCATOR;
+
+pub static IDENTICAL_MAPPING: AtomicBool = AtomicBool::new(true);
```
shrik3, line 31: hmm... what's this (ignore me if this is already in the documentations.)

qlib/kernel/Kernel.rs
```diff
@@ -42,553 +50,1314 @@ impl HostSpace {
     }
 
     pub fn LoadProcessKernel(processAddr: u64) -> i64 {
-        let mut msg = Msg::LoadProcessKernel(LoadProcessKernel {
-            processAddr: processAddr,
-        });
-
-        return HostSpace::Call(&mut msg, false) as i64;
+        if is_cc_active() {
```
shrik3, line 53: this pattern seems pretty repetitive

shrik3: IMHO this one is pretty wild ...

qkernel/src/syscalls/sys_time.rs
123abcpp: For gettimeofday in Quark, it is a hypercall that gets the time from the host directly.
In fact, currently timers are all stored in the shared memory.

qlib/kernel/arch/tee.rs
```diff
@@ -0,0 +1,116 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use core::sync::atomic::{AtomicU8, Ordering};
+use lazy_static::lazy_static;
+use crate::qlib::linux_def::MemoryDef;
+use crate::qlib::config::CCMode;
+
+lazy_static! {
+    //TODO: It should be only set once
```
shrik3, line 21: then use `once_cell` .. I think this has been stablized (also perhaps use this instead of `lazy_static` in general.

qlib/addr.rs
```diff
@@ -526,6 +530,11 @@ impl Addr {
         self.PageOffset() == 0
     }
 
+    pub fn is_huge_page_aligned(&self, _type: pagetable::HugePageType) -> bool {
```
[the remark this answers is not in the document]
123abcpp: type is a rust keyword... Cannot do that.

qlib/config.rs
```diff
@@ -138,4 +140,27 @@ pub enum CudaMemType {
     Default,
     UM, //Unified Memory
     MemPool,
-}
\ No newline at end of file
+}
+
+#[derive(Clone, Copy, Debug, PartialOrd, Ord, Eq, PartialEq, Serialize, Deserialize)]
+pub enum CCMode {
```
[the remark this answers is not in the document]
123abcpp: We have a pdf documentation above.

qlib/kernel/boot/loader.rs
```diff
@@ -127,7 +129,11 @@ impl Loader {
         };
 
         let kernel = Kernel::Init(kernelArgs);
-        *SHARESPACE.kernel.lock() = Some(kernel.clone());
+        if is_cc_active(){
+            *GUEST_KERNEL.lock() = Some(kernel.clone());
```
shrik3, line 133: What's the difference?

If I understand correctly in cc context the namespaces are very different?

qlib/kernel/fs/host/util.rs
```diff
@@ -26,7 +26,9 @@ use super::super::super::super::linux_def::*;
 use super::super::super::super::qmsg::qcall::TryOpenStruct;
 use super::super::super::kernel::time::*;
 use super::super::super::task::*;
-use super::super::super::util::cstring::*;
+use super::super::super::util::sharedcstring::*;
```
shrik3, line 29: note that cstr is stablized in std (core). (I think there was an issue on this)

qlib/kernel/fs/inode.rs
```diff
@@ -426,9 +428,9 @@ impl Inode {
     }
 
     pub fn NewTmpDirInode(task: &Task, root: &str) -> Result<Self> {
-        let mut fstat = LibcStat::default();
+        let mut fstat = Box::new_in(LibcStat::default(), GUEST_HOST_SHARED_ALLOCATOR);
```
shrik3, line 431: If I understand correctly. Even for non-cc system, such syscalls will use a kernel heap allocator (i.e. `GUEST_HOST_SHARED_ALLOCATOR`) instead of using a local variable on the stack like before. Same applies to other syscalls where a `Box::new_in` is used.

I think this is pretty heavy overhead.

qlib/kernel/boot/loader.rs
123abcpp: We store the Kernel struct in the private memory after initialization, instead of in the shared memory where might be modified by the host later.

qlib/kernel/fs/host/util.rs
123abcpp: Since cstring is an already existed structure, we can switch both of them to cstr in an individual commit.

qlib/kernel/fs/inode.rs
123abcpp: We have implemented a per vcpu allocator for shared heap as the previous vcpu allocator, don't worry.

qlib/kernel/arch/tee.rs
```diff
@@ -0,0 +1,116 @@
...
+    match mode {
+        CCMode::None => { return false; }
+        _ => { return true; }
+    }
+}
+
+/// Depending on TEE architecture, the guest physical address should be
+/// marked as shared("untrusted")/private("trusted"). The actual set/unset
+/// bit(s) on the GPA is implementation defined by the particular TEE.
+pub fn gpa_adjust_shared_bit(_address: &mut u64, _protect: bool) {
```
shrik3, line 65: if I understand correctly, you want a mut reference to the PTE, not to its address.

qlib/kernel/memmgr/pma.rs
```diff
@@ -188,8 +189,10 @@ impl PageTables {
 
             assert!(pgdEntry.is_unused());
             pudTbl = phyAddrs[3] as *mut PageTable;
+            let mut pudTbl_addr = pudTbl as u64;
```
shrik3, line 192: this doesn't look sound to me ...

qlib/kernel/quring/uring_async.rs
```diff
@@ -198,8 +205,8 @@ impl UringAsyncMgr {
 
     pub fn SetOps(&self, id: usize, ops: AsyncOps) -> UringEntry {
         // squeue::Entry {
-        *self.ops[id].lock() = ops.clone();
-
+            *self.ops[id].lock() = ops.clone();
```
shrik3, line 208: misformatted

qlib/kernel/util/sharedcstring.rs
```diff
@@ -0,0 +1,85 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use alloc::str;
+use alloc::string::String;
+use alloc::vec::Vec;
+
+use super::super::super::mem::list_allocator::GuestHostSharedAllocator;
```
shrik3, line 19: use absolute include path

qlib/linux_def.rs
```diff
@@ -17,6 +17,8 @@ use alloc::vec::Vec;
 use core::sync::atomic::Ordering;
 
 use super::super::kernel_def::*;
+use crate::qlib::mem::list_allocator::GuestHostSharedAllocator;
+use crate::GUEST_HOST_SHARED_ALLOCATOR;
```
shrik3, line 21: perhaps merge the namespaces .. this looks weird.

qkernel/src/kernel_def.rs
```diff
@@ -269,42 +339,82 @@ pub fn NewSocket(fd: i32) -> i64 {
...
+        } else {
+            let current = Task::Current().GetTaskId();
 
-        let qMsg = QMsg {
-            taskId: current,
-            globalLock: true,
-            ret: 0,
-            msg: msg,
-        };
+            let qMsg = QMsg {
```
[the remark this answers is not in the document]
123abcpp: It is a git presentation problem, not misformatted.

qlib/pagetable.rs
```diff
@@ -132,9 +136,47 @@ pub struct PageTables {
     pub hibernateLock: QMutex<()>,
 }
 
+#[derive(Copy, Clone)]
+pub enum HugePageType {
+    MB2,
```
shrik3, line 141: is this indirection really necessary?

qlib/pagetable.rs
```diff
@@ -185,7 +227,8 @@ impl PageTables {
     #[cfg(target_arch = "x86_64")]
     pub fn IsActivePagetable(&self) -> bool {
         let root = self.GetRoot();
-        return root == Self::CurrentCr3();
+        let cr3 = tee::guest_physical_address(Self::CurrentCr3());
```
shrik3, line 230: if i understand correctly, in CC context the guest registers are saved to a shared memory location, which you need this tee memory api to read out.

IMO you should modify the `CurrentCr3()` instead of this function. The indirection here is confusing.

qlib/kernel/Kernel.rs
123abcpp: Yeah this is repetitive, but I still need a place to check the current CC mode and cannot find a better way yet.

qvisor/src/qcall.rs
```diff
@@ -302,7 +302,6 @@ impl KVMVcpu {
                         .SwapOut(heapStart, heapEnd - heapStart)
                         .unwrap();
                 }
-
```
shrik3, line 305: .

qvisor/src/runc/runtime/signal_handle.rs
```diff
@@ -115,21 +117,31 @@ extern "C" fn handle_sigintAct(
...
+            let res = SignalProcess(&ROOT_CONTAINER_ID.lock().clone(), 0, signal, console);
+            debug!("SignalProcess res {:?}", res);
+        } else {
+            SendSignalVMS(signal, console);
+        }
+
     }
 }
 
+pub fn SendSignalVMS(signal: i32, console: bool) {
```
shrik3, line 130: I don't understand this name

qlib/kernel/arch/tee.rs
123abcpp: Yeah, we sometimes pass a mut reference to the PTE, but the logic is for an address.

qlib/kernel/memmgr/pma.rs
123abcpp: Why? I don't really get your idea what you want here.

qlib/addr.rs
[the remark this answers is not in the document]
shrik3: I mean, remove the leading underscore. Name can be any.

qlib/kernel/quring/uring_async.rs
123abcpp: fixed

qlib/kernel/util/sharedcstring.rs
123abcpp: fixed

qlib/kernel/memmgr/pma.rs
shrik3: perhaps I don't get the point of `gpa_adjust_shared_bit` ... since it's empty atm.

qkernel/src/kernel_def.rs
yaoxin-jing: I guess this is the overhead you need to take

qlib/linux_def.rs
123abcpp: GuestHostSharedAllocator is a struct and GUEST_HOST_SHARED_ALLOCATOR is an instance.
The only possible way is to store the instance in the same file where its struct is defined, but I would prefer to store all the allocator instances together with the GLOBAL_ALLOCATOR in the lib.rs.

qlib/pagetable.rs
123abcpp: Make it more readable.

qlib/pagetable.rs
```diff
@@ -132,9 +136,47 @@ pub struct PageTables {
...
+    }
+}
+
 impl PageTables {
+    /// We are creating mappings from the VMM side, e.g. kernel, MMIO_PAGE on arm.
+    /// If we go for not IDENTICAL_MAPPING btw Host<->Guest, addresses of allocated
+    /// tables while on host should be adjusted.
+    fn adjust_address(address: u64, to_guest: bool) -> u64 {
+        let mut mapping_offset = 0;
+        if  crate::IS_GUEST == false {
```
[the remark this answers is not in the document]
123abcpp: It is necessary for unidentical mapping during initialization. Or do you think adding the offset everywhere is better?

qvisor/src/qcall.rs
123abcpp: Caused by adding cc tag then removeing it...

qlib/addr.rs
```diff
@@ -277,7 +278,10 @@ impl PageOpts {
 
     pub fn Kernel() -> Self {
         return PageOpts(
-            PageTableFlags::VALID,
+            PageTableFlags::VALID |
+            PageTableFlags::MT_NORMAL |
+            PageTableFlags::DIRTY |
```
yaoxin-jing, line 283: purpose of adding these flags?

qlib/config.rs
```diff
@@ -138,4 +140,27 @@ pub enum CudaMemType {
...
+    None,
+    Normal,
```
yaoxin-jing, line 148: I forgot, but how do you setup cc_mode flags?

qlib/kernel/Kernel.rs
123abcpp: It is in the documentation for explaining different CCMode.
In CCMode::NormalEmu the value is set to false.

qlib/pagetable.rs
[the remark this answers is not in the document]
shrik3: I mean you have 3xif in a row without an else branch. This is nitpicking, though.

qlib/kernel/arch/tee.rs
```diff
@@ -0,0 +1,116 @@
...
+    if is_hw_tee() {
+        //
+        // Impliment according to architecture
+        //
+    }
+}
+
+/// Before the guest can reason on a GPA, the information on the IPA that
+/// regards the TEE should be removed.
```
yaoxin-jing, line 74: What dose the 'ipa' means here?

qlib/kernel/memmgr/pma.rs
123abcpp: For example in SEV SNP 
```
match get_tee_type() {
            #[cfg(feature = "snp")]
            CCMode::SevSnp => {
                if _protect {
                    *_address |= sev_snp::C_BIT_MASK.load(Ordering::Acquire);
                } else {
                    *_address &= sev_snp::C_BIT_MASK.load(Ordering::Acquire) -1;
                }
            }
```
The bit is set when converting the address to a private physical address, and unset in the opposite.

qlib/kernel/boot/loader.rs
yaoxin-jing: Because guest kenrel contain sensitive information that host shouldnt obtain via sharedspace

qvisor/src/runc/runtime/signal_handle.rs
123abcpp: Actually, it calls the signal function in vms. Do you have an idea to change it?

qlib/kernel/fs/copy_up.rs
```diff
@@ -250,7 +251,10 @@ fn copyAttributesLocked(task: &Task, upper: &mut Inode, lower: &Inode) -> Result
             continue;
         }
 
-        let value = lower.Getxattr(task, name, Xattr::XATTR_SIZE_MAX)?;
+        let value = lower
+            .Getxattr(task, name, Xattr::XATTR_SIZE_MAX)?
```
yaoxin-jing, line 255: Purpose of doing this?

qlib/addr.rs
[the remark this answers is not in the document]
123abcpp: Ok

qlib/pagetable.rs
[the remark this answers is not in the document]
123abcpp: I see, fixed.

qlib/addr.rs
123abcpp: These are aarch64 flags...

qvisor/src/runc/runtime/signal_handle.rs
shrik3: from the code it seems this is a function taken out of the original control flow. This function is only used in `signal_handle.rs` , so perhaps at least remove the `pub` to make it a module internal function, so that the naming doesn't matter much.

I was asking because it looked like a public API...

qlib/kernel/arch/tee.rs
shrik3: intermediate physical address, IIRC.

it's a guest physical address used by MMU to point to next-level paging structure. Not sure if this term is arm specific.

123abcpp: IPA stands for intermediate physical address, which is the term of guest physical address in arm.

qvisor/src/runc/runtime/signal_handle.rs
123abcpp: I see, have fixed it.

qlib/config.rs
yaoxin-jing: who set the mode in this case? tenant via k8s yaml file or cloud provider via quark config file?

123abcpp: Via quark config file.

qlib/addr.rs
yaoxin-jing: need doc for people who dont arm. Also if it is arm related code, it probably shouldnt show in this pull request at all.

shrik3: > need doc for people who dont arm. Also if it is arm related code, it probably shouldnt show in this pull request at all.

not necessary, this code is arm only `#[cfg(target_arch = "aarch64")]`

123abcpp:
> I think the patch looks good overall and should be merged to the main

Then could you please approve the PR?

PR #1359, opened 2025-03-26
fix: page table initialization flags

123abcpp:
For the first level entry, the USER_ACCESSIBLE flag should be set.


shrik3:
> For the first level entry, the USER_ACCESSIBLE flag should be set.

The kernel and user address space overlap wrt. the 1st level entry?

In which case perhaps you should call the raw APIs during initialization instead of modifying the `MapWithX` wrappers.

shrik3:
looks good to me, thanks!

shrik3 (approved):

PR #1362, opened 2025-05-19
aarch64: ensure the virtual counter value is synchronized

chl337:


shrik3:
the barrier here matters when measuring very short pieces of code, is this the case?

shrik3 (approved):

PR #1363, opened 2025-05-19
mm: modify rss field only through deticated method

chl337:


qlib/kernel/memmgr/mm.rs
```diff
@@ -504,10 +490,9 @@ impl MemoryManager {
                     mapping.lockedAS -= r.Len();
                 }
 
-                let mut pt = self.pagetable.write();
-
-                pt.pt.MUnmap(r.Start(), r.Len())?;
-                pt.curRSS -= r.Len();
+                self.pagetable.write()
+                    .pt.MUnmap(r.Start(), r.Len())?;
+                self.RemoveRssLock(&r);
```
shrik3, line 495: I think this produces redundant pagetable lock acquire/release.

PR #1364, opened 2025-06-22
fix bug: INTEL XEON 4510's fpstates capacity size is over 2 pages

yaoxin-jing:


yaoxin-jing:
> what about making this model specific i.e. conditional compiling or configurable

I don't think it worth doing that. All latest data center Intel cpu has very big fpstates capacity size. I would assume this is also the case for latest AMD data center CPUs.

shrik3:
-- if there is no regression on desktop chips (i.e. single page fpstate).

I wonder if there is a hardcoded size in the low-level code somewhere. Would you do a regression test?

PR #1370, opened 2026-06-05
unix: fix host UDS recvmsg results

### LINKED ISSUE #1369: host UDS recvmsg returns EOF or corrupt data
A Quark container can connect to a host Unix domain socket exposed through a bind mount, but reads from that socket return EOF or invalid data instead of the host reply.

This is a bug in the existing host UDS passthrough path, not a missing feature. Quark already detects a bind-mounted host socket inode, calls `HostUnixConnect`, and stores `HostUnixSocketOperations` for later send and receive calls.

The bug is in qvisor's host UDS `recvmsg` wrapper:

1. It passes the Quark hostfd to `libc::recvmsg` without translating it to the real OS fd.
2. It returns `0` for a successful `recvmsg`, so the guest sees EOF.
3. It converts `EAGAIN` into a positive byte count, so the guest can copy zero bytes as if they were data.

The fix is to translate the fd, return the real byte count, and keep errno values negative when passing them back to the guest.

Repro result before the fix:

```text
host process -> OK
runc         -> OK
runsc        -> connect-failed errno=111
quark rw     -> EOF before reply
quark ro     -> EOF before reply
```

Result after the fix:

```text
host process -> OK
runc         -> OK
runsc        -> connect-failed errno=111
quark rw     -> OK
quark ro     -> OK
```

A stricter static C client also passes under Quark for both rw and ro bind mounts. It fails if the first read returns EOF or if the reply starts with zero bytes.

russellromney:
## Summary

Fix host Unix socket `recvmsg` passthrough.

The host UDS path already connects to bind-mounted host socket inodes. The receive side was returning the wrong result to the guest.

This change:

- translates the Quark hostfd to the real OS fd before `libc::recvmsg`
- returns the actual `recvmsg` byte count
- returns errno as an error, not as a positive byte count
- uses the host UDS recv wrapper on blocking retries
- adds a static C regression client that uses `recvmsg` and can assert `SCM_RIGHTS` fd delivery with `expect-fd`

Fixes #1369.

## Test

Original bind-mounted UDS probe:

```text
host process -> OK
runc         -> OK
runsc        -> connect-failed errno=111
quark rw     -> OK
quark ro     -> OK
```

Regression client under Quark:

```text
host_uds_recv: 8 iterations ok
mode=rw ok
host_uds_recv: 8 iterations ok
mode=ro ok

The regression client also supports `expect-fd` mode for servers that send `SCM_RIGHTS`, validating control-data fd translation.
```

### DIFF
--- a/qlib/kernel/socket/unix/host_unix.rs
+++ b/qlib/kernel/socket/unix/host_unix.rs
@@ -378,11 +378,10 @@ impl HostUnixSocketOperations {
                 _ => (),
             }
 
-            res = Kernel::HostSpace::IORecvMsg(
+            res = Kernel::HostSpace::HostUnixRecvMsg(
                 self.fd,
                 &mut *msgHdr as *mut _ as u64,
                 flags | MsgType::MSG_DONTWAIT,
-                false,
             ) as i32;
         }
 
--- a/qvisor/src/vmspace/mod.rs
+++ b/qvisor/src/vmspace/mod.rs
@@ -1203,18 +1203,24 @@ impl VMSpace {
     ///////////start of network operation//////////////////////////////////////////////////////////////////
 
     pub fn HostUnixRecvMsg(fd: i32, msghdr: u64, flags: i32) -> i64 {
+        let fd = match Self::GetOsfd(fd) {
+            Some(fd) => fd,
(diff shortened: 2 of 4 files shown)


yaoxin-jing:
looks good to me

